PR #52, opened 2022-01-24
[011Y-213] Add fuzzing of encoding libraries 

Pryz:


kevinburkesegment (approved):

encoding/fuzz_test.go
```diff
@@ -0,0 +1,183 @@
...
+	})
+}
+
+func FuzzFixedLenByteArrayEncoding(f *testing.F) {
+	f.Fuzz(func(t *testing.T, size int, input []byte) {
+		fuzzEncoding(func(e encoding.Encoding) {
+			fuzzFixedLenByteArrayEncoding(t, size, input, e)
+		})
+	})
+}
```
achille-roussel, line 83: Some encodings don't support byte arrays, do we need to add fuzzing tests for other data types?

encoding/fuzz_test.go
```diff
@@ -0,0 +1,183 @@
...
+		}
+
+		if errors.Is(err, encoding.ErrInvalidArguments) {
+			return
+		}
+		t.Fatal("encode: ", err)
+	}
+
+	for i := 0; i < (len(input) / size); i++ {
+		n, err := dec.DecodeFixedLenByteArray(size, tmp)
```
achille-roussel, line 119: I think we may get a lot more leverage out of the fuzzing code if we configure the test to decode the input rather than encoding it. 

The idea is to validate that the decoders can handle malformed payloads, and don't panic or error in unexpected ways when they are exposed to invalid input.

This would be the pattern we want to use:

```go
dec := e.NewDecoder(bytes.NewReader(input))
tmp := make([]byte, length)
err := dec.DecodeFixedLenByteArray(size, tmp)
```

encoding/fuzz_test.go
```diff
@@ -0,0 +1,254 @@
...
+		fuzzEncoding(func(e encoding.Encoding) {
+			fuzzByteArrayEncoding(t, input, e)
+		})
+	})
+}
+
+func FuzzFixedLenByteArrayEncoding(f *testing.F) {
+	f.Fuzz(func(t *testing.T, size int, input []byte) {
+		fuzzEncoding(func(e encoding.Encoding) {
+			//fuzzFixedLenByteArrayEncoding(t, size, input, e)
```
kevinburkesegment, line 99: ?

encoding/notsupported.go
```diff
@@ -17,6 +17,14 @@ var (
 	// errors.Is rather than equality comparisons to test the error values
 	// returned by encoders and decoders.
 	ErrNotSupported = errors.New("encoding not supported")
+
+	// ErrInvalidArguments is an error returned when arguments passed to the
+	// encoding functions are incorrect and will lead to an expected failure.
+	//
+	// As ErrNotSupported, this error may be wrapped with specific information
```
kevinburkesegment, line 24:
```suggestion
	// As with ErrNotSupported, this error may be wrapped with specific information
```

Pryz:
Ok I think this PR is ready for review.

achille-roussel (approved):
This is looking good, can we wait for https://github.com/segmentio/parquet-go/pull/103 to be merged so we can switch from Gotip to using Go 1.18?

PR #53, opened 2022-01-28
[O11Y-196] implement parquet bloom filters

achille-roussel:
This PR adds support for generating and querying parquet bloom filters as defined by the parquet format specification in https://github.com/apache/parquet-format/blob/master/BloomFilter.md

The flavor of bloom filters implemented in parquet are "split block bloom filters" (SBBF). The main difference with classic bloom filters is in the way the filter is divided in fixed-size blocks: when a key is inserted in the filter, its filter bits are written to a single 32 bytes block rather than being dispatched across the whole filter. This design guarantees that the number of CPU cache miss remains constant (at most one) no matter how many bits per key the filter is configured to host.

## Design

I designed the APIs to abstract away most of the complexity of generating and using the filters. To create filters, the program adds a `parquet.BloomFilters` option to the `parquet.Writer` configuration, enumerating the type filters to apply to each column, for example:

```go
writer := parquet.NewWriter(output,
  parquet.BloomFilters(
    parquet.SplitBlockFilter("last_name"),
  ),
)
```

I opted for a default behavior where no bloom filters are generated, because they can quickly increase the CPU and memory footprint of the program, and applications should generally know which columns they want to create filters for.

When reading parquet files, filters can be used to determine whether a column chunk may contain a value, by accessing the filter via the `parquet.ColumnChunk.BloomFilter` method, the general pattern looks as such:

```go
lastName := rowGroup.Column(lastNameColumnIndex)
bloomFilter := lastName.BloomFilter()

if ok, err := bloomFilter.Check([]byte("Solo")); err != nil {
  ...
} else if ok {
  // the value likely exists in the column chunk
  ...
}
```

The code for the bloom filter implementation is located in the new `bloom` package.

## Implementation

In past experiences we had using bloom filters in combination with LSM trees, we observed that filter generation accounted for over half of the compute time when performing compaction of runs.

I invested in research time to figure out how to design bloom filter APIs to reduce the overhead of filter generation (inserting keys), which led me down a rabbit hole of trying to determine how much throughput was possible to achieve to have a target baseline.

It turned out that the parquet bloom filter algorithm were well designed to leverage vectorized CPU instructions, so I added both pure Go and assembly versions of those functions, the vectorized versions of the algorithms yield large improvements to the throughput of the filter compared to the code generated by the Go compiler. Here is an example taken from one of the benchmarks I added where we insert 1000 64 bits values per iteration:

```
name                old time/op   new time/op    delta
BloomFilterEncoder   13.1µs ± 3%     2.7µs ± 5%   -79.19%  (p=0.000 n=10+10)

name                old speed     new speed      delta
BloomFilterEncoder  609MB/s ± 3%  2929MB/s ± 5%  +380.89%  (p=0.000 n=10+10)
```

Before I go further, it is important to mention that parquet bloom filter are somewhat composed of two steps: to insert a key, the XXH64 hash of the value is first computed, then the resulting 64 bits key is inserted in the filter.

I first started the implementation with https://github.com/cespare/xxhash but after optimizing the bloom filter to support bulk inserts (which created more opportunities to leverage vectorization), computing the hash of individual keys accounted for most of the CPU time spent. I attempted to optimize the hash function without success, so instead I pulled the `xxhash` package into parquet-go and made modifications to introduce new APIs that support bulk computation of hashes for fixed size keys (most use cases for bloom filters would likely be on columns holding primary keys that are often integers or UUIDs). Having those specialized APIs and working on bulks of inputs made it possible to leverage vectorization to increase the throughput of XXH64 computation, which yielded a second round of optimization.

This made it clear that the higher level APIs that rely on these primitives would also need to support writing values to the filters in bulks in order to benefit from the improvements. I managed to reuse the `encoding.Encoder` interface for this purpose, which allows pages of parquet values to be written in bulk to the filter in the same way that they are written to the file.

I still need to work on documenting how to configure and use the bloom filter APIs, but overall the code should be ready for feedback!

---
Edit: I ended up removing the vectorized XXH64 code; I ran more benchmarks and the end results yield mixed returns, which doesn't seem worth the added complexity. Assembly code is hard-enough to maintain, it really needs to be worth the cost; here is the comparison to the scalar vs vectorized assembly:

```
name                    old speed      new speed      delta
MultiSum64Uint8/16B     5.55GB/s ± 1%  5.84GB/s ± 2%   +5.13%  (p=0.000 n=8+9)
MultiSum64Uint8/100B    5.76GB/s ± 1%  6.46GB/s ± 5%  +12.04%  (p=0.000 n=9+9)
MultiSum64Uint8/4KB     6.14GB/s ± 2%  6.63GB/s ± 3%   +7.98%  (p=0.000 n=9+10)
MultiSum64Uint8/10MB    5.85GB/s ± 3%  6.34GB/s ± 3%   +8.43%  (p=0.000 n=10+10)
MultiSum64Uint16/16B    3.67GB/s ± 4%  4.16GB/s ± 2%  +13.37%  (p=0.000 n=10+10)
MultiSum64Uint16/100B   4.16GB/s ± 3%  4.47GB/s ± 3%   +7.41%  (p=0.000 n=10+10)
MultiSum64Uint16/4KB    4.39GB/s ± 1%  4.61GB/s ± 2%   +4.85%  (p=0.000 n=9+9)
MultiSum64Uint16/10MB   4.26GB/s ± 4%  4.39GB/s ± 3%   +2.88%  (p=0.009 n=10+10)
MultiSum64Uint32/4B     1.94GB/s ± 5%  1.69GB/s ± 2%  -12.89%  (p=0.000 n=10+9)
MultiSum64Uint32/16B    3.80GB/s ± 3%  3.58GB/s ± 6%   -5.70%  (p=0.000 n=10+10)
MultiSum64Uint32/100B   5.25GB/s ± 3%  5.78GB/s ± 3%  +10.10%  (p=0.000 n=10+10)
MultiSum64Uint32/4KB    5.59GB/s ± 2%  6.34GB/s ± 3%  +13.25%  (p=0.000 n=10+10)
MultiSum64Uint32/10MB   5.17GB/s ± 5%  5.83GB/s ± 5%  +12.95%  (p=0.000 n=10+10)
MultiSum64Uint64/16B    2.59GB/s ± 4%  2.28GB/s ± 4%  -11.72%  (p=0.000 n=10+10)
MultiSum64Uint64/100B   3.97GB/s ± 6%  4.03GB/s ± 3%     ~     (p=0.631 n=10+10)
MultiSum64Uint64/4KB    4.39GB/s ± 3%  4.97GB/s ± 5%  +13.23%  (p=0.000 n=10+10)
MultiSum64Uint64/10MB   4.09GB/s ± 1%  4.47GB/s ± 5%   +9.15%  (p=0.000 n=8+10)
MultiSum64Uint128/16B   1.56GB/s ± 2%  1.52GB/s ± 5%   -2.92%  (p=0.021 n=8+10)
MultiSum64Uint128/100B  2.57GB/s ± 4%  2.66GB/s ± 5%   +3.64%  (p=0.010 n=9+10)
MultiSum64Uint128/4KB   3.13GB/s ± 3%  3.06GB/s ± 9%     ~     (p=0.579 n=10+10)
MultiSum64Uint128/10MB  2.84GB/s ± 3%  2.79GB/s ± 1%   -1.90%  (p=0.003 n=9+9)

name                    old hash/s     new hash/s     delta
MultiSum64Uint8/16B         694M ± 1%      730M ± 2%   +5.13%  (p=0.000 n=8+9)
MultiSum64Uint8/100B        720M ± 1%      807M ± 5%  +12.04%  (p=0.000 n=9+9)
MultiSum64Uint8/4KB         768M ± 2%      829M ± 3%   +7.98%  (p=0.000 n=9+10)
MultiSum64Uint8/10MB        731M ± 3%      792M ± 3%   +8.43%  (p=0.000 n=10+10)
MultiSum64Uint16/16B        458M ± 4%      519M ± 2%  +13.37%  (p=0.000 n=10+10)
MultiSum64Uint16/100B       520M ± 3%      559M ± 3%   +7.41%  (p=0.000 n=10+10)
MultiSum64Uint16/4KB        549M ± 1%      576M ± 2%   +4.85%  (p=0.000 n=9+9)
MultiSum64Uint16/10MB       533M ± 4%      548M ± 3%   +2.88%  (p=0.009 n=10+10)
MultiSum64Uint32/4B         243M ± 5%      211M ± 2%  -12.89%  (p=0.000 n=10+9)
MultiSum64Uint32/16B        475M ± 3%      448M ± 6%   -5.70%  (p=0.000 n=10+10)
MultiSum64Uint32/100B       657M ± 3%      723M ± 3%  +10.10%  (p=0.000 n=10+10)
MultiSum64Uint32/4KB        699M ± 2%      792M ± 3%  +13.25%  (p=0.000 n=10+10)
MultiSum64Uint32/10MB       646M ± 5%      729M ± 5%  +12.95%  (p=0.000 n=10+10)
MultiSum64Uint64/16B        323M ± 4%      285M ± 4%  -11.72%  (p=0.000 n=10+10)
MultiSum64Uint64/100B       496M ± 6%      503M ± 3%     ~     (p=0.631 n=10+10)
MultiSum64Uint64/4KB        548M ± 3%      621M ± 5%  +13.23%  (p=0.000 n=10+10)
MultiSum64Uint64/10MB       512M ± 1%      559M ± 5%   +9.15%  (p=0.000 n=8+10)
MultiSum64Uint128/16B       196M ± 2%      190M ± 5%   -2.92%  (p=0.021 n=8+10)
MultiSum64Uint128/100B      321M ± 4%      332M ± 5%   +3.64%  (p=0.010 n=9+10)
MultiSum64Uint128/4KB       391M ± 3%      382M ± 9%     ~     (p=0.579 n=10+10)
MultiSum64Uint128/10MB      355M ± 3%      349M ± 1%   -1.90%  (p=0.003 n=9+9)
```

I kept the code on a separate branch for now https://github.com/segmentio/parquet-go/tree/xxhash-avx2, I'll experiment with AVX-512 and ARM64 vectorization as follow up.


bloom.go
```diff
@@ -0,0 +1,230 @@
...
+
+// BloomFilter is an interface allowing applications to test whether a key
+// exists in a bloom filter.
+type BloomFilter interface {
+	// Implement the io.ReaderAt interface as a mechanism to allow reading the
+	// raw bits of the filter.
+	io.ReaderAt
+	// Returns the size of the bloom filter (in bytes).
+	Size() int64
+	// Tests whether the key is present in the filter.
```
kevinburkesegment, line 21: Not obvious to me why this would or could return an error.

bloom/block.go
```diff
@@ -0,0 +1,28 @@
...
+// Bytes returns b as a byte slice.
+func (b *Block) Bytes() []byte {
+	return unsafe.Slice((*byte)(unsafe.Pointer(b)), BlockSize)
+}
+
+const (
+	// BlockSize is the size of bloom filter blocks in bytes.
+	BlockSize = 32
+
+	salt0 = 0x47b6137b
```
kevinburkesegment, line 20: Where did these values come from?

bloom/block_amd64.s
```diff
@@ -0,0 +1,70 @@
...
+DATA salt+4(SB)/4, $0x44974d91
+DATA salt+8(SB)/4, $0x8824ad5b
+DATA salt+12(SB)/4, $0xa2b7289d
+DATA salt+16(SB)/4, $0x705495c7
+DATA salt+20(SB)/4, $0x2df1424b
+DATA salt+24(SB)/4, $0x9efc4947
+DATA salt+28(SB)/4, $0x5c6bfb31
+GLOBL salt(SB), RODATA|NOPTR, $32
+
+// This initial block is a SIMD implementation of the mask function declared in
```
kevinburkesegment, line 25: Did you write this yourself or borrow it from somewhere? If borrowed could we link to the original?

bloom/block_default.go
```diff
@@ -0,0 +1,65 @@
+//go:build purego && parquet.bloom.no_unroll
+
+package bloom
+
+// This file contains direct translation of the algortihms described in the
```
kevinburkesegment, line 5:
```suggestion
// This file contains direct translation of the algorithms described in the
```

bloom/block_default.go
```diff
@@ -0,0 +1,65 @@
...
+// parquet bloom filter spec:
+// https://github.com/apache/parquet-format/blob/master/BloomFilter.md
+//
+// There are no practical reasons to unable the parquet.bloom.no_unroll build
```
kevinburkesegment, line 9: Maybe "to enable"? Unable is an adjective short for "not able"

bloom/block_test.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/segmentio/parquet-go/bloom"
+)
+
+func TestBlock(t *testing.T) {
+	for i := uint64(0); i < math.MaxUint32; i = (i * 2) + 1 {
+		x := uint32(i)
+		b := bloom.Block{}
+		b.Insert(x)
+		if !b.Check(x) {
```
kevinburkesegment, line 15: Should we check x-1 or something as well

bloom/bloom_test.go
```diff
@@ -0,0 +1,37 @@
...
+
+import (
+	"math/rand"
+	"testing"
+)
+
+// Test file for internal functions of the bloom package.
+var global4x64 [4]uint64
+
+func TestFasthash(t *testing.T) {
```
kevinburkesegment, line 11: I guess another test would be that a lot of different numbers don't all hash to the same value.

bloom/filter.go
```diff
@@ -0,0 +1,105 @@
...
+}
+
+// SplitBlockFilter is an in-memory implementation of the parquet bloom filters.
+//
+// This type is useful to construct bloom filters that are later serialized
+// to a storage medium.
+type SplitBlockFilter []Block
+
+// NumSplitBlocksOf returns the number of blocks in a filter intended to hold
+// the given nuumber of values and bits of filter per value.
```
kevinburkesegment, line 34:
```suggestion
// the given number of values and bits of filter per value.
```

bloom.go
achille-roussel: The filters are serialized to storage, the API here doesn't say anything about where they are read from, so it's possible that reading the data failed and may return an error. Any practical data store should probably add a caching layer here so most of the time the filters are in memory, but this is not enforced by the library.

I'll update the documentation to reflect that.

bloom/block.go
achille-roussel: From the spec https://github.com/apache/parquet-format/blob/master/BloomFilter.md#technical-approach

bloom/block_amd64.s
achille-roussel: I wrote it.

bloom/block_default.go
achille-roussel: You are correct 👍

bloom/block_test.go
achille-roussel: It's tricky, because bloom filters may have false positives, `Check` may return `true` even if the key wasn't inserted.

Maybe I can revisit the test to do a bunch of checks that validate that the false positive rates are below the expected threshold.

bloom/filter_amd64.s
```diff
@@ -0,0 +1,129 @@
...
+// func filterInsertBulk(f []Block, x []uint64)
+TEXT ·filterInsertBulk(SB), NOSPLIT, $0-48
+    MOVQ f_base+0(FP), AX
+    MOVQ f_len+8(FP), CX
+    MOVQ x_base+24(FP), BX
+    MOVQ x_len+32(FP), DX
+    VPBROADCASTQ f_base+8(FP), Y0
+
+    // Loop initialization, SI holds the current index in `x`, DI is the number
+    // of elements in `x` rounded down the nearest multiple of 4.
```
kevinburkesegment, line 44:
```suggestion
    // of elements in `x` rounded down to the nearest multiple of 4.
```

bloom/filter_amd64.s
```diff
@@ -0,0 +1,129 @@
...
+    XORQ SI, SI
+    MOVQ DX, DI
+    SHRQ $2, DI
+    SHLQ $2, DI
+loop4x64:
+    CMPQ SI, DI
+    JAE loop
+
+    // The masks and indexes for 4 input hashes are computed in each loop
+    // iteration. The hashes are loaded in Y1 so we can used vector instructions
```
kevinburkesegment, line 54:
```suggestion
    // iteration. The hashes are loaded in Y1 so we can use vector instructions
```

bloom/filter_amd64.s
```diff
@@ -0,0 +1,129 @@
...
+    generateMask(Y4, Y8)
+    generateMask(Y5, Y9)
+
+    // The next block of instructions move indexes from the vector to general
+    // purpose registers in order to use them as offsets when applying the mask
+    // to the filter.
+    extract4x64(Y1, X1, X10, R8, R9, R10, R11)
+
+    // Apply masks to the filter; this operation is sensitive to aliasing, the
+    // blocks overlap the CPU has to serialize the reads and writes, which has
```
kevinburkesegment, line 76: > the blocks overlap the CPU has to serialize the reads and writes

This doesn't really scan. Maybe "if the blocks overlap the CPU has to serialize" ?

bloom/filter_amd64.s
```diff
@@ -0,0 +1,129 @@
...
+    // a measurable impact on throughput. This would be frequent for small bloom
+    // filters which may have only a few blocks, the probability of seeing
+    // overlapping blocks on large filters should be small enough to make this
+    // a non-issue tho.
```
kevinburkesegment, line 80:
```suggestion
    // a non-issue though.
```

bloom/filter_amd64.s
```diff
@@ -0,0 +1,129 @@
...
+    applyMask(Y6, (AX)(R8*1))
+    applyMask(Y7, (AX)(R9*1))
+    applyMask(Y8, (AX)(R10*1))
+    applyMask(Y9, (AX)(R11*1))
+
+    ADDQ $4, SI
+    JMP loop4x64
+loop:
+    // Compute trailing elements in `x` if the length was not a multiple of 4.
+    // This is the same algorthim as the one in the loop4x64 section, working
```
kevinburkesegment, line 90:
```suggestion
    // This is the same algorithm as the one in the loop4x64 section, working
```

bloom/bloom_test.go
achille-roussel: I think it would make sense if we were "inventing" the algorithm, but in this case it's copied from the parquet specs https://github.com/apache/parquet-format/blob/master/BloomFilter.md#technical-approach, it's an internal API, and just a one-liner, so probably a bit overkill to invest more time in testing it.

bloom/filter_test.go
```diff
@@ -0,0 +1,135 @@
...
+			}
+		}
+	})
+}
+
+func TestSplitBlockFilterBug1(t *testing.T) {
+	h := [1]uint64{0b1000101001000001001001111000000100011011001000011110011100110000}
+	f := make(bloom.SplitBlockFilter, 1)
+	f.InsertBulk(h[:])
+	if !f.Check(h[0]) {
```
kevinburkesegment, line 84: check h[0]+1, h[0]-1 too?

bloom/hash.go
```diff
@@ -0,0 +1,49 @@
...
+	// Compute hashes of the array of fixed size values passed as arguments,
+	// returning the number of hashes written to the destination buffer.
+	MultiSum64Uint8(dst []uint64, src []uint8) int
+	MultiSum64Uint16(dst []uint64, src []uint16) int
+	MultiSum64Uint32(dst []uint64, src []uint32) int
+	MultiSum64Uint64(dst []uint64, src []uint64) int
+	MultiSum64Uint128(dst []uint64, src [][16]byte) int
+}
+
+// XXH64 is an implementation fo the Hash interface using the XXH64 algortihm.
```
kevinburkesegment, line 20:
```suggestion
// XXH64 is an implementation fo the Hash interface using the XXH64 algorithm.
```

bloom/xxhash/sum64uint_amd64.go
```diff
@@ -0,0 +1,43 @@
...
+// MultiSum64Uint128/4KB  1.84GB/s ± 2%  3.11GB/s ± 4%  +68.70%  (p=0.000 n=9+10)
+//
+// name                   old hash/s     new hash/s     delta
+// MultiSum64Uint8/4KB        617M ± 2%      852M ± 5%  +38.00%  (p=0.000 n=10+10)
+// MultiSum64Uint16/4KB       431M ± 2%      579M ± 4%  +34.56%  (p=0.000 n=10+10)
+// MultiSum64Uint32/4KB       605M ± 2%      799M ± 4%  +31.94%  (p=0.000 n=10+10)
+// MultiSum64Uint64/4KB       471M ± 2%      618M ± 2%  +31.14%  (p=0.000 n=9+10)
+// MultiSum64Uint128/4KB      231M ± 2%      389M ± 4%  +68.70%  (p=0.000 n=9+10)
+//
+// The benchmarks measure the throughput of hashes produces, as a rate of values
```
kevinburkesegment, line 27:
```suggestion
// The benchmarks measure the throughput of hashes produced, as a rate of values
```

bloom/xxhash/xxhash.go
```diff
@@ -0,0 +1,55 @@
...
+)
+
+const (
+	prime1 uint64 = 0x9E3779B185EBCA87
+	prime2 uint64 = 0xC2B2AE3D27D4EB4F
+	prime3 uint64 = 0x165667B19E3779F9
+	prime4 uint64 = 0x85EBCA77C2B2AE63
+	prime5 uint64 = 0x27D4EB2F165667C5
+	// Precomputed operations because the compiler otherwise complains that the
+	// results overflow 64 bits integers.
```
kevinburkesegment, line 17:
```suggestion
	// results overflow 64 bit integers.
```

column_chunk.go
```diff
@@ -25,6 +25,13 @@ type ColumnChunk interface {
 	// If the column chunk does not have a page index, the methods return nil.
 	ColumnIndex() ColumnIndex
 	OffsetIndex() OffsetIndex
+	BloomFilter() BloomFilter
+
+	// Returns the number of volumnes in the column chunk.
```
kevinburkesegment, line 30:
```suggestion
	// Returns the number of values in the column chunk.
```

Values? Columns? Volumes?

kevinburkesegment (approved):
Okay - I tried my best to review everything, I didn't really check for correctness of the code or the assembly bits, I did try to check the tests covered the useful cases though. 

It would be good to have a top-level example of how to configure this and maybe more docs about _why_ you'd want to use bloom filters in your Parquet application.

bloom/filter_test.go
achille-roussel: I documented why this test was added instead 👍

bloom/block_test.go
achille-roussel: I added your suggestion in be1a59b

achille-roussel:
Thanks for the thorough review @kevinburkesegment !

PR #58, opened 2022-02-02
add test demonstrating how to read a subset of columns

achille-roussel:
This PR adds a test I wrote as an example to show how to read a subset of the columns.

### DIFF
--- a/reader_test.go
+++ b/reader_test.go
@@ -337,3 +337,40 @@ func BenchmarkReader(b *testing.B) {
 		})
 	}
 }
+
+func TestReaderReadSubset(t *testing.T) {
+	type Point3D struct { X, Y, Z int64 }
+	type Point2D struct { X, Y int64 }
+
+	f := func(points3D []Point3D) bool {
+		if len(points3D) == 0 {
+			return true
+		}
+		buf := new(bytes.Buffer)
+		err := writeParquetFile(buf, makeRows(points3D))
+		if err != nil {
+			t.Error(err)
+			return false
+		}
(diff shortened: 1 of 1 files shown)


reader_test.go
```diff
@@ -337,3 +337,40 @@ func BenchmarkReader(b *testing.B) {
...
+		if len(points3D) == 0 {
+			return true
+		}
+		buf := new(bytes.Buffer)
+		err := writeParquetFile(buf, makeRows(points3D))
+		if err != nil {
+			t.Error(err)
+			return false
+		}
+		reader := parquet.NewReader(bytes.NewReader(buf.Bytes()))
```
kevinburkesegment, line 355: A `bytes.Buffer` implements `Read()` doesn't it? I don't think you need to call `.Bytes()` here

reader_test.go
```diff
@@ -337,3 +337,40 @@ func BenchmarkReader(b *testing.B) {
 		})
 	}
 }
+
+func TestReaderReadSubset(t *testing.T) {
+	type Point3D struct { X, Y, Z int64 }
+	type Point2D struct { X, Y int64 }
+
+	f := func(points3D []Point3D) bool {
```
kevinburkesegment, line 345:
```go
// In this example we'll write 3 columns to the file - X, Y, and Z, but we'll only read out the X and Y columns.
// Returns true if all writes and reads were successful, and false otherwise.
```

kevinburkesegment (approved):

reader_test.go
achille-roussel: Yes, but the signature is `parquet.NewReader(io.ReaderAt)` :)

PR #59, opened 2022-02-07
[O11Y-197] optimize min/max algorithms on x86 with AVX-512 support

achille-roussel:
I've started this a couple weeks ago but didn't find the time to wrap up the work until today, so just submitting now.

This PR revisits the implementations of a couple of internal algorithms that we use when creating parquet files, using vectorized AVX-512 instructions to improve the throughput of those operations on parquet columns.

The assembly code can be complex in places, but the work being done is very simple in general, we look for min/max values in an arrays of primitive types. I've made sure to invest in thorough testing to minimize the risk of seeing regressions introduced by those optimizations.

For what it's worth, those are the measured improvements on some of the benchmarks I added in the PR:

```
name                             old time/op    new time/op      delta
MinMaxBool/4KiB                    1.23µs ± 0%      0.04µs ± 0%    -96.79%  (p=0.000 n=8+10)
MinMaxBool/256KiB                  77.6µs ± 0%       3.3µs ± 0%    -95.73%  (p=0.000 n=10+9)
MinMaxBool/2048KiB                  624µs ± 0%        64µs ± 8%    -89.78%  (p=0.000 n=9+10)
MinMaxInt32/4KiB                    640ns ± 0%        44ns ± 0%    -93.11%  (p=0.000 n=9+10)
MinMaxInt32/256KiB                 40.7µs ± 0%       3.3µs ± 0%    -91.84%  (p=0.000 n=10+9)
MinMaxInt32/2048KiB                 327µs ± 0%        60µs ± 7%    -81.57%  (p=0.000 n=10+10)
MinMaxInt64/4KiB                    331ns ± 0%        45ns ± 0%    -86.40%  (p=0.000 n=9+8)
MinMaxInt64/256KiB                 20.9µs ± 0%       3.4µs ± 0%    -83.87%  (p=0.000 n=10+9)
MinMaxInt64/2048KiB                 169µs ± 0%        62µs ± 8%    -63.28%  (p=0.000 n=9+10)
MinMaxUint32/4KiB                   905ns ± 0%        44ns ± 0%    -95.14%  (p=0.000 n=9+8)
MinMaxUint32/256KiB                58.1µs ± 0%       3.3µs ± 0%    -94.31%  (p=0.000 n=10+9)
MinMaxUint32/2048KiB                467µs ± 0%        59µs ± 8%    -87.32%  (p=0.000 n=10+10)
[...]

PR #60, opened 2022-02-08
[O11Y-197] optimize add/sub algorithms on x86 with AVX-512 support

achille-roussel:
This PR is based on #59, it adds two more optimizations for add/sub routines. Those are used in the implementation of delta encoding.


kevinburkesegment:
How are you deciding which functions to rewrite in assembler?

kevinburkesegment (approved):

achille-roussel:
> How are you deciding which functions to rewrite in assembler?

Pretty much anything that's on the write path and operates on array of values is a great candidate:

- For the use cases we have, we are going to ingest lots of data which is why I'm focusing on optimizing writes
- Arrays of fixed size values map well to SIMD instructions

For this particular PR, the add/sub functions are used in DELTA_BINARY_PACKED encoding for example.

Usually the workflow I follows was to first write tests to validate the behavior of the routines (valuable work whether we choose to optimize or not), write benchmarks to have a baseline to compare to, then write a simple version of a SIMD code to estimate what potential gains are and if it's worth spending more time on it.

PR #63, opened 2022-02-08
[O11Y-197] fix AVX512BW extension requirements

achille-roussel:
Follow up from #59, this PR fixes the instructions used and feature checks to minimize the reliance on AVX512BW.

### DIFF
--- a/internal/bits/bits_amd64.go
+++ b/internal/bits/bits_amd64.go
@@ -9,6 +9,20 @@ var hasAVX512 = cpu.X86.HasAVX512 &&
 var hasAVX512MinMaxBool = hasAVX512 &&
 	cpu.X86.HasAVX512VPOPCNTDQ
 
+// The use AVX-512 instructions in the minBOol algorithm relies operations
+// that are avilable in the AVX512BW extension:
+// * VPCMPUB
+// * KMOVQ
+var hasAVX512MinBool = hasAVX512 &&
+	cpu.X86.HasAVX512BW
+
+// The use AVX-512 instructions in the maxBOol algorithm relies operations
+// that are avilable in the AVX512BW extension:
+// * VPCMPUB
+// * KMOVQ
+var hasAVX512MaxBool = hasAVX512 &&
+	cpu.X86.HasAVX512BW
+
 // The use AVX-512 instructions in the countByte algorithm relies operations
 // that are avilable in the AVX512BW extension:
 // * VPCMPUB
--- a/internal/bits/bits_amd64.go
+++ b/internal/bits/bits_amd64.go
@@ -8,3 +8,13 @@ var hasAVX512 = cpu.X86.HasAVX512 &&
 
 var hasAVX512MinMaxBool = hasAVX512 &&
 	cpu.X86.HasAVX512VPOPCNTDQ
+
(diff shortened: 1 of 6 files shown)


internal/bits/bits_amd64.go
```diff
@@ -8,3 +8,27 @@ var hasAVX512 = cpu.X86.HasAVX512 &&
 
 var hasAVX512MinMaxBool = hasAVX512 &&
 	cpu.X86.HasAVX512VPOPCNTDQ
+
+// The use AVX-512 instructions in the minBOol algorithm relies operations
```
kevinburkesegment, line 12:
```suggestion
// These use AVX-512 instructions in the minBool algorithm relies operations
```

internal/bits/bits_amd64.go
```diff
@@ -8,3 +8,27 @@ var hasAVX512 = cpu.X86.HasAVX512 &&
...
+// that are avilable in the AVX512BW extension:
+// * VPCMPUB
+// * KMOVQ
+var hasAVX512MinBool = hasAVX512 &&
+	cpu.X86.HasAVX512BW
+
+// The use AVX-512 instructions in the maxBOol algorithm relies operations
```
kevinburkesegment, line 19:
```suggestion
// These use AVX-512 instructions in the maxBool algorithm relies operations
```

internal/bits/bits_amd64.go
```diff
@@ -8,3 +8,27 @@ var hasAVX512 = cpu.X86.HasAVX512 &&
...
+// that are avilable in the AVX512BW extension:
+// * VPCMPUB
+// * KMOVQ
+var hasAVX512MaxBool = hasAVX512 &&
+	cpu.X86.HasAVX512BW
+
+// The use AVX-512 instructions in the countByte algorithm relies operations
```
kevinburkesegment, line 26:
```suggestion
// These use AVX-512 instructions in the countByte algorithm relies operations
```

internal/bits/count_amd64.s
```diff
@@ -31,7 +31,7 @@ TEXT ·countByte(SB), NOSPLIT, $0-40
     SHRQ $8, DX
     SHLQ $8, DX
     ADDQ AX, DX
-    VPBROADCASTB value+24(FP), Z0
+    VPBROADCASTB BX, Z0
```
kevinburkesegment, line 34: I don't see where we are setting BX but just going to assume one of the instructions manipulates it and you know what you're doing here haha.

kevinburkesegment, line 34: How did you catch this? Would be neat if we had an automated checker for instructions like this.

kevinburkesegment (approved):

internal/bits/count_amd64.s
achille-roussel: Actually it was picked up by CI in some ways, the Github runners seem to not always execute on the same CPU types, some runs crashed because the code was executing instructions unknown to the CPU (because the features weren't checked properly).

I agree tho, I think AVO auto-generates comments that say which CPU features are required by a function, but something more advanced that could run in CI would be great (e.g. looking at all instructions and answering they are not executed if CPU features are not checked).

PR #64, opened 2022-02-15
[O11Y-218] add interfaces to read columns of values from parquet pages

achille-roussel:
This PR adds APIs to read arrays of column values. To experiment with different approaches we discussed, I have implemented two models based on `io.Reader`/`io.Writer` and interfaces with stronger typing. Here are examples of how programs would use these:

### reading columns of typed values
```go
switch p := page.(type) {
case parquet.Int64Reader:
    n, err = p.ReadInt64s(values)
}
```

### writing columns of typed values
```go
column := buffer.Column(0).(parquet.Int64Writer)
n, err := column.WriteInt64s(values)
if err != nil {
    ...
}
```

An similarly with implementations of `io.Reader`/`io.Writer`, in places where those are better interfaces to integrate with the program.

This is definitely just a first step to test the design, the implementation focuses only on in-memory pages, what is NOT in this PR:
- a solution for reading typed arrays from optional or repeated columns, we need to figure out how to extract information about null values or lengths of repetitions, for now programs should still fallback to `parquet.ValueReader` and `parquet.ValueWriter` in those cases.

- implementation of the interfaces for pages read from parquet files, this is a more complex work that I would like to keep as a follow up if we feel good about the approach.

- documentation on those APIs

### DIFF
--- a/column_buffer.go
+++ b/column_buffer.go
@@ -2,11 +2,14 @@ package parquet
 
 import (
 	"bytes"
+	"fmt"
 	"io"
 	"sort"
 
 	"github.com/segmentio/parquet-go/deprecated"
 	"github.com/segmentio/parquet-go/encoding"
+	"github.com/segmentio/parquet-go/encoding/plain"
+	"github.com/segmentio/parquet-go/internal/bits"
 )
 
 // ColumnBuffer is an interface representing columns of a row group.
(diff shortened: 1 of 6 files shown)


pelletier (approved):
The typed interface seems more intuitive to me. When first seeing the `Read` and `Write` functions I immediately wondered how the data had to be laid out, what happened if I provided an odd-sized `b`, etc.

achille-roussel:
Thanks for the feedback! I'll make sure to properly document the expectations when using the `io` interfaces 👍

PR #66, opened 2022-02-15
[O11Y-217] always use int64 to represent row indexes

achille-roussel:
This PR unifies external APIs to always use `int64` instead of `int` for row indexes.

### DIFF
--- a/column_buffer.go
+++ b/column_buffer.go
@@ -293,10 +293,10 @@ func (col *optionalColumnBuffer) WriteValues(values []Value) (n int, err error) 
 
 func (col *optionalColumnBuffer) WriteRow(row Row) error {
 	if len(row) == 0 {
-		return errRowHasTooFewValues(len(row))
+		return errRowHasTooFewValues(int64(len(row)))
 	}
 	if len(row) > 1 {
-		return errRowHasTooManyValues(len(row))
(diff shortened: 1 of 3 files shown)


pelletier:
What's the motivation behind this change?

achille-roussel:
Go makes no strong guarantees on the capacity of a `int` value, it's "at least 32 bits", and there may be more than 2 billion rows to in a parquet file (or set of merged files).

pelletier:
I see a bunch of casting like `int64(len(...))`. I'd expect that value to be at most the maximum value of `int`. When `int` is only 32 bits, isn't the program still not going to be able to handle sets of more than 2B values?

achille-roussel:
> I see a bunch of casting like int64(len(...)). I'd expect that value to be at most the maximum value of int. When int is only 32 bits, isn't the program still not going to be able to handle sets of more than 2B values?

On 32 bits systems, I would expect that the maximum memory addressable by a program would be 2GiB, so it wouldn't be possible to create slices which exceed the available memory.

kevinburkesegment (approved):

pelletier:
> On 32 bits systems, I would expect that the maximum memory addressable by a program would be 2GiB, so it wouldn't be possible to create slices which exceed the available memory.

I'd expect that too. Is that a change to future-proof the API then?

Issue #65, opened 2022-02-15
Investigate misuse of AVX-512 CPU feature checks


achille-roussel:
It seems the checks we have to validate whether AVX-512 instructions are available isn't correct in some functions:

```
SIGILL: illegal instruction
PC=0x4edc1f m=3 sigcode=2
instruction bytes: 0x62 0xe2 0xfd 0x48 0x7c 0xdf 0x62 0x61 0xfe 0x48 0x6f 0x3d 0xf1 0xf1 0x6 0x0

goroutine 71 [running]:
github.com/segmentio/parquet-go/internal/bits.maxBE[128](https://github.com/segmentio/parquet-go/runs/5206204595?check_suite_focus=true#step:6:128)({0xc0000af800, 0x100, 0x100})
	/home/runner/work/parquet-go/parquet-go/internal/bits/max_amd64.s:467 +0x3f fp=0xc00008d3c0 sp=0xc00008d3b8 pc=0x4edc1f
github.com/segmentio/parquet-go/internal/bits.MaxFixedLenByteArray(0x0, {0xc0000af800, 0x0, 0x18})
	/home/runner/work/parquet-go/parquet-go/internal/bits/max.go:21 +0x32 fp=0xc00008d400 sp=0xc00008d3c0 pc=0x4eb7d2
github.com/segmentio/parquet-go/internal/bits_test.TestMaxFixedLenByteArray16.func1({0xc0000af800, 0x10, 0x1})
	/home/runner/work/parquet-go/parquet-go/internal/bits/max_test.go:176 +0xce fp=0xc00008d498 sp=0xc00008d400 pc=0x4faaae
runtime.call32(0xc00006a720, 0x538288, 0x0, 0x0, 0x0, 0x18, 0xc00008d9d8)
	/opt/hostedtoolcache/go/1.17.6/x64/src/runtime/asm_amd64.s:626 +0x49 fp=0xc00008d4c8 sp=0xc00008d498 pc=0x461dc9
runtime.reflectcall(0x50a7e0, 0xc0002ff938, 0x4, 0x52fdad, 0x0, 0x12, 0x50a7e0)
	<autogenerated>:1 +0x3c fp=0xc00008d508 sp=0xc00008d4c8 pc=0x465f3c
reflect.Value.call({0x50ec20, 0x538288, 0x4f183b}, {0x52cc57, 0x4}, {0xc00008dea8, 0x1, 0x10})
[...]

PR #67, opened 2022-02-15
debug issue with misuse of CPU features

achille-roussel:
Fixes https://github.com/segmentio/parquet-go/issues/65

The code was using `CMPQ` instead of `CMPB`, reading 7 extra bytes beyond the limits of the boolean variable that controlled CPU features for the min/max BE128 routines.

### DIFF
--- a/internal/bits/max_amd64.s
+++ b/internal/bits/max_amd64.s
@@ -458,7 +458,7 @@ TEXT ·maxBE128(SB), NOSPLIT, $-48
     CMPQ DX, $256
     JB loop
 
-    CMPQ ·hasAVX512MinMaxBE128(SB), $0
+    CMPB ·hasAVX512MinMaxBE128(SB), $0
     JB loop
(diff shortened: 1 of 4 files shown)


internal/bits/max_amd64.s
```diff
@@ -458,7 +458,7 @@ TEXT ·maxBE128(SB), NOSPLIT, $-48
     CMPQ DX, $256
     JB loop
 
-    CMPQ ·hasAVX512MinMaxBE128(SB), $0
```
kevinburkesegment, line 461: Huh I'm not able to find much reading about the CMPQ instruction at all.

kevinburkesegment (approved):

internal/bits/max_amd64.s
achille-roussel: https://www.felixcloutier.com/x86/cmp

PR #68, opened 2022-02-16
Increase column limit

achille-roussel:
Following up from our discussion with the Tempo and Parca folks, the limit on 127 columns appeared to be a bit too low for some use cases.

This PR modifies the package to support parquet schemas with up to 32K columns, which should be plenty for any practical use case of parquet.

I also made changes to the public API to always accept and return `int` values for repetition levels, definition levels, and column indexes, to give us the liberty to modify internal limits in the future without breaking code that depend on parquet-go. Finally, I added checks at API boundaries to ensure that conversions from `int` to smaller types do not cause overflows, and trigger panics with descriptive error messages if they do.

Let me know if you have any questions!

### DIFF
--- a/type.go
+++ b/type.go
@@ -14,7 +14,7 @@ import (
 
 // Kind is an enumeration type representing the physical types supported by the
 // parquet type system.
-type Kind int16
+type Kind int8
 
 const (
 	Boolean           Kind = Kind(format.Boolean)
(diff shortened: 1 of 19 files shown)


type.go
```diff
@@ -14,7 +14,7 @@ import (
 
 // Kind is an enumeration type representing the physical types supported by the
 // parquet type system.
-type Kind int16
+type Kind int8
```
kevinburkesegment, line 17: Just want to confirm you wanted to make this change, since it's not mentioned in the PR description and goes the opposite direction of the change described in the PR description

kevinburkesegment (approved):
See comments but this looks OK to me.

type.go
achille-roussel: Yes, I forgot to mention it, just adjusted this type to match the size we use in `parquet.Value.kind`.

It's true this could be made `int`, it's a bit ambiguous since it's a type definition so applications have to do a conversion anyways if they use the value in contexts where it comes from a different integer type... 🤷  not sure what the _right_ thing to do is.

achille-roussel:
Thanks for the quick review!

brancz:
Woah from conversation to merged PR in ~3h. Thank you so much!

PR #69, opened 2022-02-16
fix README example of using parquet.CopyRows

achille-roussel:
Reported by @annanay25


pelletier (approved):
Would be nice to be able to test the readme the same way examples in godoc are tested.

kevinburkesegment:
Was just about to leave the same comment @pelletier haha.

achille-roussel:
Some day!

PR #71, opened 2022-02-17
Reader read subset of columns

achille-roussel:
This PR fixes reports that we were missing an obvious optimization to avoid loading pages when reading rows with subsets of columns.


Pryz (approved):

achille-roussel:
Thanks for the review!

PR #72, opened 2022-02-18
[O11Y-218] add parquet.ColumnReader API to allow reading columns of typed values from parquet files

achille-roussel:
This PR is a follow up to #64, it modifies the internal implementation of reading file pages to expose types that implement the `parquet.[Type]Reader` interfaces as a mechanism to read columns of typed values from parquet files (prior to this change, it was only possible to do so from in-memory buffers).

I still need to add documentations describing how to leverage those interfaces, but the code changes should be ready to review.


file.go
```diff
@@ -912,30 +912,40 @@ func (s *filePageValueReaderState) init(columnType Type, column *Column, codec f
...
 	s.page.decoder = makeDecoder(s.page.decoder, s.page.encoding, pageEncoding, pageData)
-
-	s.repetitions.encoding = repetitionLevelEncoding
-	s.definitions.encoding = definitionLevelEncoding
 	s.page.encoding = pageEncoding
 
+	maxRepetitionLevel := column.maxRepetitionLevel
+	maxDefinitionLevel := column.maxDefinitionLevel
+	hasLevels := maxRepetitionLevel > 0 || maxDefinitionLevel > 0
+	if hasLevels {
```
achille-roussel, line 922: I noticed there was an easy optimization here, we can avoid creating the repetition and definition level decoders when dealing with required columns.

column_reader.go
```diff
@@ -0,0 +1,872 @@
...
+		bufferSize:  bufferSize,
+		columnIndex: ^columnIndex,
+	}
+}
+
+func (r *int64ColumnReader) Type() Type { return r.typ }
+
+func (r *int64ColumnReader) Column() int { return int(^r.columnIndex) }
+
+func (r *int64ColumnReader) ReadInt64s(values []int64) (n int, err error) {
```
brancz, line 428: I believe this is meant so we can more easily convert parquet to arrow. In that case I think it would be very helpful, if the functions that return the native types such as this one, read a validity boolean mask as well so the read result can be passed directly to the respective type's `AppendValues` function. [`Int64Builder.AppendValues`](https://pkg.go.dev/github.com/apache/arrow/go/arrow@v0.0.0-20211112161151-bc219186db40/array#Int64Builder.AppendValues) in this case.

achille-roussel: Arrow integration is one of the intentions here indeed. 

I've been thinking about how to introduce these bitmaps of null values, I'm currently edging towards having a second set of interfaces for optional columns, with methods that support loading both the values and bitmaps, for example:

```go
type OptionalInt32Reader interface {
  ReadOptionalInt32s(nulls bitmap, values []int32) (int, error)
}
```

Separating the methods can give stronger type checking guarantees and avoids pushing the complexity of dealing with the bitmaps in code that doesn't need it... on the other hand, it's a lot more code to manage in the library. 

Let me know if you have suggestions

brancz: I like the additional set of interfaces!

README.md
```diff
@@ -353,6 +357,130 @@ for i, n := 0, file.NumRowGroups(); i < n; i++ {
 }
 ```
 
+## Optimizations
+
+The following sections describe common optimizations techniques supported by the
```
kevinburkesegment, line 362:
```suggestion
The following sections describe common optimization techniques supported by the
```

README.md
```diff
@@ -353,6 +357,130 @@ for i, n := 0, file.NumRowGroups(); i < n; i++ {
...
+        ...
+    default:
+        values := make([]parquet.Value, page.NumValues())
+        _, err := page.ReadValues(values)
+        ...
+    }
+}
+```
+
+Reading arrays of typed values is often preferrable when performing aggregations
```
kevinburkesegment, line 403:
```suggestion
Reading arrays of typed values is often preferable when performing aggregations
```

README.md
```diff
@@ -353,6 +357,130 @@ for i, n := 0, file.NumRowGroups(); i < n; i++ {
...
+on the values as this model offers a more compact representation of the values
+in memory, and pairs well with the use of optimizations like SIMD vectorization.
+
+### Optimizing Writes
+
+Applications that deal with columnar storage are sometimes designed to work with
+columnar data throughput the abstraction layers; it then becomes possible to
```
kevinburkesegment, line 410:
```suggestion
columnar data throughout the abstraction layers; it then becomes possible to
```

I think?

README.md
```diff
@@ -353,6 +357,130 @@ for i, n := 0, file.NumRowGroups(); i < n; i++ {
...
+write columns of values directly instead of reconstructing rows from the column
+values. The package offers two main mechanisms to satisfy those use cases:
+
+#### Writing Columns of Typed Arrays
```
kevinburkesegment, line 414:
```suggestion
#### 1. Writing Columns of Typed Arrays
```

README.md
```diff
@@ -353,6 +357,130 @@ for i, n := 0, file.NumRowGroups(); i < n; i++ {
...
+    return nil
+}
+```
+
+The latter is more efficient as it does not require boxing the input into an
+intermediary array of `parquet.Value`. However, it may not always be the right
+model depending on the situation, sometimes the generic abstraction can be a
+more expressive model.
+
+#### Implementing parquet.RowGroup
```
kevinburkesegment, line 470:
```suggestion
#### 2. Implementing parquet.RowGroup
```

README.md
```diff
@@ -353,6 +357,130 @@ for i, n := 0, file.NumRowGroups(); i < n; i++ {
...
+
+Programs that need full control over the construction of row groups can choose
+to provide their own implementation of the `parquet.RowGroup` interface, which
+includes defining implementations of `parquet.ColumnChunk` and `parquet.Page`
+to expose column values of the row group.
+
+This model can be preferrable when the underlying storage or in-memory
```
kevinburkesegment, line 477:
```suggestion
This model can be preferable when the underlying storage or in-memory
```

column_reader.go
```diff
@@ -0,0 +1,872 @@
...
+	r.offset = 0
+	r.count = 0
+}
+
+// The types below are implementations of the ColumnReader interface for each
+// primitive type supported by parquet.
+//
+// The readers use an in-memory intermediary buffer to support decoding arrays
+// of values from the underlying decoder, which are then boxed into the []Value
+// buffer passed to ReadValues. When the program converts type checks the
```
kevinburkesegment, line 253: Two verbs in a row here, I think you want either "converts" or "type checks"

column_reader.go
```diff
@@ -0,0 +1,872 @@
...
+// readers for more specific interfaces (e.g. parquet.Int32Reader), the values
+// can be decoded directly from the underlying decoder, there is no need for
```
kevinburkesegment, line 255:
```suggestion
// can be decoded directly from the underlying decoder. There is no need for
```

column_reader.go
```diff
@@ -0,0 +1,872 @@
...
+func (r *booleanColumnReader) Column() int { return int(^r.columnIndex) }
+
+func (r *booleanColumnReader) ReadBooleans(values []bool) (n int, err error) {
+	if r.offset < len(r.buffer) {
+		n = copy(values, r.buffer[r.offset:])
+		r.offset += n
+		values = values[n:]
+	}
+	if r.decoder == nil {
+		err = io.EOF
```
kevinburkesegment, line 287: return n, io.EOF?

Then you can dedent

page.go
```diff
@@ -45,6 +45,12 @@ type Page interface {
 	Size() int64
 
 	// Returns a reader exposing the values contained in the page.
+	//
+	// Depending on the underlying implementation, the returned reader may
+	// support reading array of typed Go values by implementing interfaces
```
kevinburkesegment, line 50:
```suggestion
	// support reading an array of typed Go values by implementing interfaces
```

kevinburkesegment (approved):

achille-roussel:
Thanks everyone for the reviews!

PR #75, opened 2022-02-19
change parquet.BloomFilter.Check to accept a parquet.Value

achille-roussel:
This PR modifies the `parquet.BloomFilter` interface to accept a `parquet.Value` instead of a `[]byte`. The intent is to reduce reliance on a particular encoding, which results in more readable code in general in my opinion. It also creates better compile and runtime type checking, since the API does not accept any arbitrary byte slice anymore.

**Before**
```go
b := [8]byte{}
binary.LittleEndian.PutUint64(b[:], v)

if ok, err := bloomFilter.Check(b[:]); err != nil {
    ...
} else if ok {
    ...
}
```

**After**
```go
x := parquet.ValueOf(v)

if ok, err := bloomFilter.Check(x); err != nil {
    ...
} else if ok {
    ...
}
```

### DIFF
--- a/bloom.go
+++ b/bloom.go
@@ -22,15 +22,12 @@ type BloomFilter interface {
 
 	// Tests whether the given value is present in the filter.
 	//
-	// The parquet value is expected to be PLAIN-encoded, as it would be by
-	// calling the parquet.Value.Bytes method.
-	//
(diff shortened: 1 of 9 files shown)


pelletier (approved):
Nice!

PR #76, opened 2022-02-22
Expose rowGroupRowReader to be used by custom RowGroup implementations

brancz:
I have a custom implementation of the `RowGroup` interface and am just missing a reader, and the standard reader here works just fine. I do wonder if we should introduce a separate interface that has everything except the `Rows` method for this as it feels like an incorrect abstraction when we pass something that according to the interface can already build a row reader. Happy to introduce an interface for that is identical to `RowGroup` except that it doesn't have the `Rows` method, but unsure about the name. Maybe `RawRowGroup`?

@achille-roussel

### DIFF
--- a/row_group.go
+++ b/row_group.go
@@ -222,6 +222,10 @@ func (r *rowGroup) SortingColumns() []SortingColumn { return r.sorting }
 func (r *rowGroup) Schema() *Schema                 { return r.schema }
 func (r *rowGroup) Rows() Rows                      { return &rowGroupRowReader{rowGroup: r} }
 
+func NewRowGroupRowReader(rowGroup RowGroup) Rows {
+	return &rowGroupRowReader{rowGroup: rowGroup}
+}
+
 type rowGroupRowReader struct {
 	rowGroup RowGroup
 	schema   *Schema


achille-roussel (approved):
Thanks for submitting the PR! The approach you suggest here sounds like a good first step 👍 

I agree that since the construction of most row group readers can be satisfied by this model, it's worth asking whether we should have the `Rows` method on the `RowGroup` interface. The devil's in the details and the _"most"_.

Some constructs are not possible if the only model we have are generic row readers agnostic of the underlying properties of the row groups, which is what motivated the addition of the `Rows` method: by delegating the construction to the row group we make it possible for each row group to provide row readers that suit their particular characteristics. Here are examples of this:

- https://github.com/segmentio/parquet-go/blob/da19029f5c6c381ca12e335e6ae66750bb522bf5/convert.go#L548-L554

- https://github.com/segmentio/parquet-go/blob/da19029f5c6c381ca12e335e6ae66750bb522bf5/merge.go#L19-L23

It's possible that there are better solutions to these constraints, for now let's merge this change and experiment further? A better design may become more obvious along the way.

All that being said, thanks a lot for the feedback and contribution!

brancz:
I think we can keep it this way, for now, I don't think anyone would notice if they don't write their own RowGroup implementation.

PR #78, opened 2022-02-22
Set columnIndex when reading back a value from a dictionary

### LINKED ISSUE #77: Value column index incorrectly maintained in dictionaries
Once again, it's totally possible that I'm misusing the format and or library, but I believe that when using a `parquet.Group` to build a schema instead of using the `parquet.SchemaOf` function and build a row manually (correctly specifying the index of each value in the row), and then read that row back, then the values have incorrect index values, meaning when the values are used in further steps, for example writing it to another buffer, then the row is detected to be faulty.

Here is some example code that demonstrates it:

```go
package main

import (
	"fmt"
	"io"
	"testing"

	"github.com/segmentio/parquet-go"
)

func TestExample(t *testing.T) {
	g := parquet.Group{}
	g["a"] = parquet.Encoded(parquet.String(), &parquet.RLEDictionary)
	g["b"] = parquet.Encoded(parquet.String(), &parquet.RLEDictionary)
	g["c"] = parquet.Encoded(parquet.String(), &parquet.RLEDictionary)
	s := parquet.NewSchema("test", g)
	buf := parquet.NewBuffer(s)

	row := parquet.Row{
		parquet.ValueOf("value1").Level(0, 0, 0),
		parquet.ValueOf("value2").Level(0, 0, 1),
		parquet.ValueOf("value3").Level(0, 0, 2),
	}
	fmt.Println("write row")
	for _, v := range row {
		fmt.Printf("%#+v\n", v)
	}
	err := buf.WriteRow(row)
	if err != nil {
		t.Fatal(err)
	}

	buf2 := parquet.NewBuffer(s)
	rows := buf.Rows()
	for {
		row, err := rows.ReadRow(nil)
		if err == io.EOF {
			break
		}
		if err != nil && err != io.EOF {
			panic(err)
		}
		fmt.Println("write row")
		for _, v := range row {
			fmt.Printf("%#+v\n", v)
		}
		err = buf2.WriteRow(row)
		if err != nil {
			panic(err)
		}
	}
}
```

And the output:

```
$ go test
write row
C:0 D:0 R:0 V:value1
C:1 D:0 R:0 V:value2
C:2 D:0 R:0 V:value3
write row
C:-1 D:0 R:0 V:value1
C:-1 D:0 R:0 V:value2
C:-1 D:0 R:0 V:value3
--- FAIL: TestExample (0.00s)
panic: runtime error: index out of range [-1] [recovered]
        panic: runtime error: index out of range [-1]

goroutine 4 [running]:
testing.tRunner.func1.2({0x102f22020, 0x1400051e000})
        /opt/homebrew/Cellar/go/1.17.6/libexec/src/testing/testing.go:1209 +0x258
testing.tRunner.func1(0x1400011c9c0)
        /opt/homebrew/Cellar/go/1.17.6/libexec/src/testing/testing.go:1212 +0x284
panic({0x102f22020, 0x1400051e000})
        /opt/homebrew/Cellar/go/1.17.6/libexec/src/runtime/panic.go:1038 +0x21c
github.com/segmentio/parquet-go.(*Buffer).WriteRow(0x1400015e5b0, {0x1400007c2a0, 0x3, 0x4})
        /Users/brancz/src/github.com/segmentio/parquet-go/buffer.go:197 +0x2f8
github.com/parca-dev/parquetbug.TestExample(0x1400011c9c0)
        /Users/brancz/src/github.com/parca-dev/parquetbug/main_test.go:47 +0x8c8
testing.tRunner(0x1400011c9c0, 0x102f46750)
        /opt/homebrew/Cellar/go/1.17.6/libexec/src/testing/testing.go:1259 +0xfc
created by testing.(*T).Run
        /opt/homebrew/Cellar/go/1.17.6/libexec/src/testing/testing.go:1306 +0x328
exit status 2
FAIL    github.com/parca-dev/parquetbug 0.149s
```

brancz:
Fixes #77 

I'm not 100% sure if this makes sense nor do I like very much that this is setting the unexported struct value `columnIndex` of the `Value` struct, but this appears to fix what I was observing in #77.

### DIFF
--- a/dictionary.go
+++ b/dictionary.go
@@ -762,14 +762,18 @@ type indexedPageReader struct {
 }
 
 func (r *indexedPageReader) ReadValues(values []Value) (n int, err error) {
+	var v Value
 	for n < len(values) && r.offset < len(r.page.values) {
-		values[n] = r.page.dict.Index(int(r.page.values[r.offset]))
+		v = r.page.dict.Index(int(r.page.values[r.offset]))
+		v.columnIndex = r.page.columnIndex
+		values[n] = v
 		r.offset++
 		n++
 	}
 	if r.offset == len(r.page.values) {
 		err = io.EOF
 	}
+
 	return n, err
 }
 


achille-roussel (approved):
Once again I believe you are correct, the page readers are expected to set the column index on values they return; this is done here for boolean pages for example: https://github.com/segmentio/parquet-go/blob/main/page.go#L660-L671

The column index not being set on dictionary pages was a miss, thanks for the fix!

PR #79, opened 2022-02-23
Optional dict column buffer page panic

brancz:
I don't have a fix for this yet, but I noticed that when a column is both optional and a dictionary, after writing it in-memory into a buffer and then attempting to write it into a persistent parquet file (here simulated using a `bytes.Buffer`) then it panics.

This PR so far adds a test that panics, but I haven't found a fix yet, but in case it is obvious I wanted to open the PR already.

// edit

I believe I have found the culprit, and also why it only happens sometimes, it looks like the `WriteRow` and `WriteValues` functions did not produce identical buffers. I believe the patch now fixes that.

### DIFF
--- a/column_buffer.go
+++ b/column_buffer.go
@@ -258,26 +258,33 @@ func (col *optionalColumnBuffer) WriteValues(values []Value) (n int, err error) 
 	rowIndex := int32(col.base.Len())
 
 	for n < len(values) {
+		// Collect index range of contiguous null values, from i to n. If this
+		// for loop exhausts the values, all remaining if statements and for
+		// loops will be no-ops and the loop will terminate.
 		i := n
 		for n < len(values) && values[n].definitionLevel != col.maxDefinitionLevel {
 			n++
 		}
 
+		// Write the contiguous null values up until the first non-null value
+		// obtained in the for loop above.
 		for _, v := range values[i:n] {
-			col.rows = append(col.rows, rowIndex)
+			col.rows = append(col.rows, -1)
 			col.definitionLevels = append(col.definitionLevels, v.definitionLevel)
-			rowIndex++
 		}
 
+		// Collect index range of contiguous non-null values, from i to n.
 		i = n
 		for n < len(values) && values[n].definitionLevel == col.maxDefinitionLevel {
 			n++
 		}
 
+		// As long as i < n we have non-null values still to write. It is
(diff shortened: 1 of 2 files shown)


column_buffer.go
```diff
@@ -258,26 +258,31 @@ func (col *optionalColumnBuffer) WriteValues(values []Value) (n int, err error)
...
 		i := n
 		for n < len(values) && values[n].definitionLevel != col.maxDefinitionLevel {
 			n++
 		}
 
+		// Write the contiguous null values up until the first non-null value
+		// obtained in the for loop above.
 		for _, v := range values[i:n] {
-			col.rows = append(col.rows, rowIndex)
+			col.rows = append(col.rows, -1)
```
brancz, line 270: This now replicates what the `optionalColumnBuffer.WriteRow` function does.

column_buffer.go
```diff
@@ -258,26 +258,31 @@ func (col *optionalColumnBuffer) WriteValues(values []Value) (n int, err error)
...
 			n++
 		}
 
+		// As long as i < n we have non-null values still to write. It is
+		// possible that we just exhausted the input values in which casn i == n
+		// and the outer for loop will terminate.
 		if i < n {
 			count, err := col.base.WriteValues(values[i:n])
 			col.definitionLevels = appendLevel(col.definitionLevels, col.maxDefinitionLevel, count)
-			n += count
```
brancz, line 280: I couldn't understand whether there was some use to this, but I believe it was mistakenly added even though the above for loop already incremented `n`.

column_buffer.go
```diff
@@ -258,26 +258,31 @@ func (col *optionalColumnBuffer) WriteValues(values []Value) (n int, err error)
...
 		}
 
+		// Collect index range of contiguous non-null values, from i to n.
 		i = n
 		for n < len(values) && values[n].definitionLevel == col.maxDefinitionLevel {
 			n++
 		}
 
+		// As long as i < n we have non-null values still to write. It is
+		// possible that we just exhausted the input values in which casn i == n
```
kevinburkesegment, line 281:
```suggestion
		// possible that we just exhausted the input values in which case i == n
```

brancz:
I believe all comments on this PR have been addressed. Let me know if there is anything else!

achille-roussel (approved):
Thanks for the fixes 🙌

PR #86, opened 2022-03-04
Preserve correct column index in null values

brancz:
Before this patch, all null values read from an optional column had the `-1` column index, while non-null values had their correct, expected index.

### DIFF
--- a/page.go
+++ b/page.go
@@ -353,10 +353,14 @@ func (r *optionalPageReader) ReadValues(values []Value) (n int, err error) {
 		r.values = r.page.base.Values()
 	}
 	maxDefinitionLevel := r.page.maxDefinitionLevel
+	columnIndex := ^int16(r.page.Column())
 
 	for n < len(values) && r.offset < len(r.page.definitionLevels) {
 		for n < len(values) && r.offset < len(r.page.definitionLevels) && r.page.definitionLevels[r.offset] != maxDefinitionLevel {
-			values[n] = Value{definitionLevel: r.page.definitionLevels[r.offset]}
+			values[n] = Value{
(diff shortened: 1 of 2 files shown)


page.go
```diff
@@ -356,7 +356,10 @@ func (r *optionalPageReader) ReadValues(values []Value) (n int, err error) {
 
 	for n < len(values) && r.offset < len(r.page.definitionLevels) {
 		for n < len(values) && r.offset < len(r.page.definitionLevels) && r.page.definitionLevels[r.offset] != maxDefinitionLevel {
-			values[n] = Value{definitionLevel: r.page.definitionLevels[r.offset]}
+			values[n] = Value{
+				definitionLevel: r.page.definitionLevels[r.offset],
+				columnIndex:     ^int16(r.page.Column()),
```
achille-roussel, line 361: We could move this computation out of the loop, the `Column` call is done on an interface so the compiler won't be able to know the value doesn't change across iterations.

achille-roussel (approved):
Looking great 🙌  thanks for the fix!

page.go
brancz: Do you prefer it to be in the reader creation or just at the beginning of the function? I'm slightly leaning towards the first just because it prevents more unnecessary calls.

achille-roussel: I think having it at the beginning of the function is the right balance between not doing unnecessary work and keeping the code simple. 

Caching it in the reader _could_ be the most efficient but it requires somewhat keeping more _global_ state which always make it harder to reason about the code. I'd go for the local variable computed before the loop, which should already amortize the cost to the point that it shouldn't be measurable on any CPU profiles.

achille-roussel (approved):
Thanks for the fix!

PR #87, opened 2022-03-04
Expose sort functions

brancz:
I'm building things around the `RowGroup`, essentially a construct that always ensures that all data is sorted. In order to do that with values read from a RowGroup externally I need the sort functions to be usable from the outside of this package.

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -202,7 +202,7 @@ type rowGroupCursor interface {
 
 type columnSortFunc struct {
 	columnIndex int16
-	compare     sortFunc
+	compare     SortFunc
 }
 
 type bufferedRowGroupCursor struct {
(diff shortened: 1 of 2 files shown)


merge.go
```diff
@@ -266,9 +266,9 @@ var (
 	//_ RowWriterTo         = (*mergedRowGroupRowReader)(nil)
 )
 
-type sortFunc func(a, b []Value) int
+type SortFunc func(a, b []Value) int
 
-func sortFuncOf(t Type, maxDefinitionLevel, maxRepetitionLevel int8, descending, nullsFirst bool) (sort sortFunc) {
+func SortFuncOf(t Type, maxDefinitionLevel, maxRepetitionLevel int8, descending, nullsFirst bool) (sort SortFunc) {
```
achille-roussel, line 271: I'm a bit concerned about exposing this API as-is, it doesn't really follow good practices of being hard to misuse: for example, there are two boolean parameters which say nothing about what they are at the call site (e.g. `parquet.SortFuncOf(typ, d, r, false, false)`), the consecutive arguments of same types also make it easy to mistakenly swap the parameters and introduce bugs that compilers or linters cannot detect.

Let me circle back on this with a proposal later today.

brancz: Yeah, that makes perfect sense. Let me know if you come up with an interface you like!

achille-roussel: > the consecutive arguments of same types also make it easy to mistakenly swap the parameters and introduce bugs that compilers or linters cannot detect.

As a matter of fact, there was actually a bug in the implementation where the repetition and definition levels were inverted in the call to `sortFunc` 🤦  I should have known to apply my own advises!

This should be fixed in https://github.com/segmentio/parquet-go/pull/93

achille-roussel:
Closing in favor of https://github.com/segmentio/parquet-go/pull/93

PR #88, opened 2022-03-07
Prevent attempting to reorder optional rows when unchanged

brancz:
In our system, a higher-level layer already ensures that all buffers are sorted, so we do a bunch of unnecessary work attempting to reorder a column whose order hasn't changed. This essentially copies the approach that the repeated buffers took.

### DIFF
--- a/column_buffer.go
+++ b/column_buffer.go
@@ -184,6 +184,13 @@ func (col *optionalColumnBuffer) Pages() Pages {
 }
 
 func (col *optionalColumnBuffer) Page() BufferedPage {
+	if !optionalRowsHaveBeenReordered(col.rows) {
+		// No need for any cyclic sorting if the rows have not been reordered.
+		// This case is also important because the cyclic sorting modifies the
+		// buffer which makes it unsafe to read the buffer concurrently.
+		return newOptionalPage(col.base.Page(), col.maxDefinitionLevel, col.definitionLevels)
+	}
+
 	numNulls := countLevelsNotEqual(col.definitionLevels, col.maxDefinitionLevel)
 	numValues := len(col.rows) - numNulls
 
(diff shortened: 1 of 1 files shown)


column_buffer.go
```diff
@@ -644,7 +646,21 @@ func (col *repeatedColumnBuffer) Values() ValueReader {
 	return &repeatedPageReader{page: col.Page().(*repeatedPage)}
 }
 
-func rowsHaveBeenReordered(rows []region) bool {
+func optionalRowsHaveBeenReordered(rows []int32) bool {
+	i := int32(0)
+	for _, row := range rows {
```
kevinburkesegment, line 651:
```go
for i := 0; i < len(rows); i++ {
    if i != rows[i]
}
```

?

column_buffer.go
```diff
@@ -184,36 +184,38 @@ func (col *optionalColumnBuffer) Pages() Pages {
 }
 
 func (col *optionalColumnBuffer) Page() BufferedPage {
-	numNulls := countLevelsNotEqual(col.definitionLevels, col.maxDefinitionLevel)
-	numValues := len(col.rows) - numNulls
+	if optionalRowsHaveBeenReordered(col.rows) {
```
kevinburkesegment, line 187: Make the fast case more straightforward? I don't know

```go
if !optionalRowsHaveBeenReordered(col.rows) {
     	return newOptionalPage(col.base.Page(), col.maxDefinitionLevel, col.definitionLevels)
} 
...
```

brancz: Sure no problem, I was just replicating the same way the equivalent repeated buffer does it. Will do!

column_buffer.go
brancz: That's not quite the same, we can only increase `i` if we actually see a non-null row.

brancz: I think this is not super obvious so I added some code comments that should clarify it for future readers.

brancz:
I believe all comments on this PR have been addressed. Let me know if there is anything else!

achille-roussel (approved):
Looking good 👍

PR #89, opened 2022-03-10
Add test reproducing unexpected fixed size binary dict error

brancz:
I don't have a fix for this yet, but I noticed that dicts of repeated UUIDs don't appear to work as expected.

This test currently fails with:

```
--- FAIL: TestWriterRepeatedUUIDDict (0.00s)
    writer_test.go:559: reading dictionary of column "list": reading parquet dictionary of fixed-length binary values of size 16: length of fixed byte array is not a multiple of its size: size=16 length=1
FAIL
```

### DIFF
--- a/writer_test.go
+++ b/writer_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 	"testing/quick"
 
+	"github.com/google/uuid"
 	"github.com/hexops/gotextdiff"
 	"github.com/hexops/gotextdiff/myers"
 	"github.com/hexops/gotextdiff/span"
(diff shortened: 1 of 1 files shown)


kevinburkesegment:
I can take this one, it might help me get more familiar with what the code is doing.

kevinburkesegment:
Currently troubleshooting this - it looks like we're writing the right UUID data to the file/buffer, looking at the read path now.

kevinburkesegment:
@brancz I submitted #113 which includes the fix - mind if I close?

brancz:
Of course! Awesome!

Issue #90, opened 2022-03-11
Data races in tests


brancz:
I'm not sure yet whether this is about the test setup, or the actual parquet-go code.

```
==================
WARNING: DATA RACE
Write at 0x00c0052645d8 by goroutine 21:
  bytes.(*Reader).Reset()
      /opt/homebrew/Cellar/go/1.17.6/libexec/src/bytes/reader.go:157 +0x69c
  github.com/segmentio/parquet-go.(*filePages).readPage()
      /Users/brancz/src/github.com/segmentio/parquet-go/file.go:533 +0x64c
  github.com/segmentio/parquet-go.(*filePages).ReadPage()
      /Users/brancz/src/github.com/segmentio/parquet-go/file.go:585 +0xfc
  github.com/segmentio/parquet-go.(*columnPages).ReadPage()
      /Users/brancz/src/github.com/segmentio/parquet-go/column.go:146 +0xec
  github.com/segmentio/parquet-go_test.forEachPage()
      /Users/brancz/src/github.com/segmentio/parquet-go/parquet_test.go:41 +0x38
  github.com/segmentio/parquet-go_test.forEachColumnPage.func1()
      /Users/brancz/src/github.com/segmentio/parquet-go/parquet_test.go:74 +0x98
  github.com/segmentio/parquet-go_test.forEachLeafColumn()
      /Users/brancz/src/github.com/segmentio/parquet-go/parquet_test.go:27 +0x6c
  github.com/segmentio/parquet-go_test.forEachLeafColumn()
      /Users/brancz/src/github.com/segmentio/parquet-go/parquet_test.go:31 +0xe0
  github.com/segmentio/parquet-go_test.forEachColumnPage()
      /Users/brancz/src/github.com/segmentio/parquet-go/parquet_test.go:73 +0x50
  github.com/segmentio/parquet-go_test.forEachColumnValue()
      /Users/brancz/src/github.com/segmentio/parquet-go/parquet_test.go:79 +0x50
  github.com/segmentio/parquet-go_test.scanParquetValues()
      /Users/brancz/src/github.com/segmentio/parquet-go/writer_test.go:38 +0x3c
  github.com/segmentio/parquet-go_test.scanParquetFile()
      /Users/brancz/src/github.com/segmentio/parquet-go/writer_test.go:34 +0xdc
  github.com/segmentio/parquet-go_test.generateParquetFile()
[...]

PR #92, opened 2022-03-11
Run tests with race detector on

brancz:
I think https://github.com/segmentio/parquet-go/issues/90 could have been caught earlier if the race detector was always on when running tests. The tests in this repo run quite fast so this wasn't an inconvenient increase, but happy to discuss.


achille-roussel (approved):
I didn't have it enabled because we weren't using any concurrency in the parquet-go code, but I forgot about dependencies :D 

This seems like a good use of a compute time, if it grows too slow we could consider using the race detector on the compress package only.

PR #93, opened 2022-03-14
add parquet.SortFunc

achille-roussel:
This PR is a proposal for exposing new APIs to generate column sorting functions. The changes intend to address the comments I left on the original proposal from https://github.com/segmentio/parquet-go/pull/87

Let me know if you have any feedback!


brancz:
I like this a lot. Happy to go exactly with this!

sort.go
```diff
@@ -0,0 +1,182 @@
...
+
+// Apply applies options to c.
+func (c *SortConfig) Apply(options ...SortOption) {
+	for _, opt := range options {
+		opt.ConfigureSort(c)
+	}
+}
+
+// ConfigureSort satisfies the SortOption interface.
+func (c *SortConfig) ConfigureSort(config *SortConfig) {
```
kevinburkesegment, line 29: I've made this comment before, but the ergonomics of having a value and then calling a function on it with another value of the same type, only to copy it, just feels very odd... 

Why do you want/need to pass it in? Can't you just have `c` pass a `SortConfig` to the caller?

Like I would maybe name this 

```go
func (c *SortConfig) Clone() *SortConfig {
    
}
```

achille-roussel: As documented on the `SortConfig` type, the intent of this method is to allow passing the configuration as option to the constructor function, I'm not sure how we could do that with the `Clone` method, can you share more details of how you would do this?

kevinburkesegment: To be honest I'd probably just make `SortFuncOf` accept a single argument, SortConfig, that's configured however people want things to be sorted...

kevinburkesegment:
Agreed, I don't think it's a big thing and happy to just merge as is, but this is our chance to get the API right, so wanted to make a point of it.

sort.go
achille-roussel: Maybe we can open an issue to discuss the config management strategy we would prefer to adopt in the library, it's a broad topic with lots of material available both online and in internal code bases.

kevinburkesegment:
Understood. I think the Option... style might make more sense when we're configuring more than just fields on a struct?

achille-roussel:
> this is our chance to get the API right, so wanted to make a point of it.

For now I'm going to keep it consistent with the other configuration APIs, but I'll start a discussion about it in a Github issue.

PR #103, opened 2022-03-17
refactor primitive parquet types to leverage Go 1.18 generics

achille-roussel:
Follow up to #98, this PR modifies more of the internal types to use Go 1.18 generics in order to reduce the amount of code that we have to maintain.


dictionary_go18.go
```diff
@@ -0,0 +1,92 @@
...
+		n, err := d.class.decode(decoder, d.values[len(d.values):cap(d.values)])
+		if n > 0 {
+			d.values = d.values[:len(d.values)+n]
+		}
+
+		if err != nil {
+			if err == io.EOF {
+				err = nil
+			} else {
+				err = fmt.Errorf("reading parquet dictionary of int64 values: %w", err)
```
Pryz, line 75: should we replace int64 by the type of T here ?

Pryz (approved):

PR #104, opened 2022-03-17
Add failing test for checking bloomfilter of a dict encoded column

brancz:
I don't have a fix yet, but I noticed that bloomfilter checks don't seem to work as expected on dictionary encoded columns. This adds a test to show the unexpected behavior.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -762,6 +762,12 @@ func (c *writerColumn) flushFilterPages() error {
 		if c.page.filter == nil {
 			c.page.filter = c.newBloomFilterEncoder(numValues)
 		}
+
+		// If there is a dictionary, we need to only write the dictionary.
+		if dict := c.dictionary; dict != nil {
+			return dict.WriteTo(c.page.filter)
+		}
+
 		for _, page := range c.filter {
 			if err := page.WriteTo(c.page.filter); err != nil {
 				return err
--- a/writer_test.go
+++ b/writer_test.go
@@ -527,3 +527,41 @@ func TestWriterGenerateBloomFilters(t *testing.T) {
 		t.Error(err)
 	}
 }
+
+func TestBloomFilterForDict(t *testing.T) {
+	type testStruct struct {
+		A string `parquet:"a,dict"`
+	}
+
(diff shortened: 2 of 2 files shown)


writer.go
```diff
@@ -763,6 +763,11 @@ func (c *writerColumn) flushFilterPages() error {
 			c.page.filter = c.newBloomFilterEncoder(numValues)
 		}
 		for _, page := range c.filter {
+			dict := page.Dictionary()
+			if dict != nil {
+				dict.WriteTo(c.page.filter)
+				continue
+			}
```
achille-roussel, line 770: It should be safe to assume that if one page has a dictionary, all pages do and will share the dictionary since they are part of the same column chunk here? 

Could we lookup the dictionary once and have two loops instead?

writer.go
```diff
@@ -762,9 +762,23 @@ func (c *writerColumn) flushFilterPages() error {
...
+		// It's safe to assume that if the first page has a dictionary all will.
+		if c.filter[0].Dictionary() == nil {
+			for _, page := range c.filter {
+				if err := page.WriteTo(c.page.filter); err != nil {
+					return err
+				}
+			}
+		} else {
+			for _, page := range c.filter {
+				page.Dictionary().WriteTo(c.page.filter)
```
achille-roussel, line 781: How about this shorter form, would it do the same?
```suggestion
		// It's safe to assume that if the first page has a dictionary all will.
		if dict := c.dictionary; dict == nil {
			for _, page := range c.filter {
				if err := page.WriteTo(c.page.filter); err != nil {
					return err
				}
			}
		} else {
			for _, page := range c.filter {
				dict.WriteTo(c.page.filter)
```
We should probably check the error returned by `WriteTo` as well in the second loop.

achille-roussel: 🤔 

Looking more at the code, do we need a loop in the second case? The dictionary will be the same for all pages, maybe we can only write it once to the filter, next calls to `WriteTo` would be no-ops.

brancz: yeah I realized the same, pushed the change!

achille-roussel (approved):
🔥

PR #105, opened 2022-03-17
Handle empty file.

Pryz:
These changes make so that we don't panic
when handling empty files. Not sure in which
case this is going to happen.


writer_test.go
```diff
@@ -565,3 +570,28 @@ func TestBloomFilterForDict(t *testing.T) {
...
+	fs, err := tmp.Stat()
+	if err != nil {
+		t.Error(err)
+	}
+
+	_, err = parquet.OpenFile(tmp, fs.Size())
+	if !errors.Is(err, parquet.ErrMissingRootColumn) {
+		t.Errorf("expected error %s, got %s", parquet.ErrMissingRootColumn, err)
+	}
+}
```
achille-roussel, line 597: In this test there is no schema defined because none were passed in when constructing the writer, and no rows were written so it assumes an empty schema.

Could we add a test where the writer is given a schema as well?

row_group.go
```diff
@@ -424,9 +424,14 @@ type emptyRowGroup struct {
 
 func newEmptyRowGroup(schema *Schema) *emptyRowGroup {
 	g := &emptyRowGroup{
-		schema:  schema,
-		columns: make([]emptyColumnChunk, numLeafColumnsOf(schema)),
+		schema: schema,
 	}
+	if schema == nil {
+		return g
+	}
```
achille-roussel, line 431: Allowing the schema to be `nil` is going to be error prone and we may have to edit documentation that stated that the schema should never be `nil` in some cases.

Do you think we could have a placeholder schema value representing the empty schema? (kind of like `struct{}` in Go). That way we can continue to require it to never be `nil`, and panic if we ever access a `nil` schema.

writer_test.go
Pryz: Added a test. Working nicely and creating valid empty file.

writer_test.go
```diff
@@ -565,3 +571,85 @@ func TestBloomFilterForDict(t *testing.T) {
 		t.Error("bloom filter should have contained 'test'")
 	}
 }
+
+func TestEmptyFile(t *testing.T) {
+	f := func(t *testing.T, schema *parquet.Schema) {
```
kevinburkesegment, line 576: `t.Helper()`?

writer_test.go
```diff
@@ -565,3 +571,85 @@ func TestBloomFilterForDict(t *testing.T) {
...
+	}
+
+	for _, test := range tests {
+		t.Run(test.scenario, func(t *testing.T) {
+			f(t, test.schema)
+		})
+	}
+}
+
+func TestFileWithNoSchema(t *testing.T) {
```
kevinburkesegment, line 622: Wait... the PR says we _shouldn't_ panic here?

file.go
```diff
@@ -209,7 +209,13 @@ func (f *File) readPageIndex(section *bufferedSectionReader, decoder *thrift.Dec
...
 
 // RowGroup returns the row group at the given index in f.
-func (f *File) RowGroup(i int) RowGroup { return &f.rowGroups[i] }
+// Returns nil is the file is empty.
+func (f *File) RowGroup(i int) RowGroup {
+	if len(f.rowGroups) == 0 {
+		return nil
+	}
+	return &f.rowGroups[i]
+}
```
achille-roussel, line 218: Should we leave this panic instead? It seems similar to indexing into an empty slice, the program should check for the number of row groups instead?

reader.go
```diff
@@ -73,8 +73,13 @@ func NewReader(input io.ReaderAt, options ...ReaderOption) *Reader {
...
 
-	column := f.Root()
-	schema := NewSchema(column.Name(), column)
+	isEmpty := f.Size() == 0
+
+	var schema *Schema
+	if !isEmpty {
+		column := f.Root()
+		schema = NewSchema(column.Name(), column)
+	}
```
achille-roussel, line 82: If I'm reading this correctly, it's meant to allow zero-size files? Should we allow it considering zero-size files aren't valid parquet files?

writer.go
```diff
@@ -143,7 +141,13 @@ func (w *Writer) Write(row interface{}) error {
 //
 // The row is expected to contain values for each column of the writer's schema,
 // in the order produced by the parquet.(*Schema).Deconstruct method.
-func (w *Writer) WriteRow(row Row) error { return w.writer.WriteRow(row) }
+func (w *Writer) WriteRow(row Row) error {
+	if w.schema == nil {
+		return ErrRowGroupSchemaMissing
+	}
+
```
achille-roussel, line 148: Maybe a shorter version?
```suggestion
```

writer.go
```diff
@@ -153,6 +157,10 @@ func (w *Writer) WriteRow(row Row) error { return w.writer.WriteRow(row) }
 // The content of the row group is flushed to the writer; after the method
 // returns successfully, the row group will be empty and in ready to be reused.
 func (w *Writer) WriteRowGroup(rowGroup RowGroup) (int64, error) {
+	if rowGroup.NumRows() == 0 {
+		return 0, nil
+	}
```
achille-roussel, line 162: The row group may have no rows, but it may carry a schema and a list of sorting columns that the programs would intent to write to the file. The check here would prevent this.

Maybe we could add a tests to verify these conditions are supported?

reader.go
Pryz: Should we panic instead ?

writer_test.go
Pryz: Sorry, how is t.Helper will help here ?

kevinburkesegment: Point out the actual failing test, but I guess this is alright as is.

writer_test.go
Pryz: fixed

reader.go
Pryz: switched to a panic

Pryz:
Closing for now. This PR is not relevant anymore due to all the underlying changes.

PR #106, opened 2022-03-18
Optimize looking up min/max values of pages with dictionaries

achille-roussel:
This PR adds a benchmark to highlight a performance bottleneck that was reported when computing the min/max bounds of pages with dictionaries, and proposes introducing a new `parquet.Dictionary.Bounds` method optimizing the operation.

After analyzing CPU profiles, it appeared that the excessive compute footprint was mostly driven by the cost of abstractions: multiple indirect method calls through interface and deep call stacks in a tight loop over large data sets add up quickly.

To amortize the cost of abstractions, I am adding a `Bounds(indexes []int32) (min, max Value)` method to the `parquet.Dictionary` interface, that the indexed pages delegate the computation of min/max values to, passing the list of dictionary indexes that belong to the page as argument. With this model, we cross abstraction boundaries only once per bound lookup, instead of once for each comparison, which amortizes the compute cost.

Here is a comparison of the benchmark I added highlighting the impact of the change:
```
name                             old time/op    new time/op    delta
DictionaryPageBounds/BOOLEAN       58.9µs ± 1%    27.6µs ± 1%  -53.22%  (p=0.000 n=10+9)
DictionaryPageBounds/INT32         65.6µs ± 0%    38.4µs ± 1%  -41.49%  (p=0.000 n=10+10)
DictionaryPageBounds/INT64         69.9µs ± 1%    42.0µs ± 1%  -39.89%  (p=0.000 n=9+10)
DictionaryPageBounds/INT96          127µs ± 1%      64µs ± 1%  -49.41%  (p=0.000 n=8+10)
DictionaryPageBounds/FLOAT         73.0µs ± 1%    46.0µs ± 1%  -37.02%  (p=0.000 n=10+10)
DictionaryPageBounds/DOUBLE        74.9µs ± 0%    47.7µs ± 0%  -36.38%  (p=0.000 n=10+6)
DictionaryPageBounds/BYTE_ARRAY    70.1µs ± 1%    41.0µs ± 1%  -41.51%  (p=0.000 n=10+10)

name                             old alloc/op   new alloc/op   delta
DictionaryPageBounds/BOOLEAN        331kB ± 0%     331kB ± 0%   -0.01%  (p=0.000 n=10+10)
DictionaryPageBounds/INT32          303kB ± 0%     303kB ± 0%   -0.01%  (p=0.000 n=10+10)
DictionaryPageBounds/INT64          393kB ± 0%     393kB ± 0%   -0.01%  (p=0.000 n=10+10)
[...]

PR #107, opened 2022-03-18
Optimize inserting values in dictionaries

achille-roussel:
Following up from #106, this PR applies the same optimization strategy to inserting values in dictionary pages.

Deltas from the parent branch:
```
name                             old time/op    new time/op    delta
DictionaryPageBounds/BOOLEAN       27.6µs ± 1%    21.9µs ± 0%  -20.75%  (p=0.000 n=9+9)
DictionaryPageBounds/INT32         38.4µs ± 1%    34.1µs ± 2%  -11.11%  (p=0.000 n=10+10)
DictionaryPageBounds/INT64         42.0µs ± 1%    37.6µs ± 1%  -10.39%  (p=0.000 n=10+10)
DictionaryPageBounds/INT96         64.3µs ± 1%    57.1µs ± 0%  -11.18%  (p=0.000 n=10+9)
DictionaryPageBounds/FLOAT         46.0µs ± 1%    38.7µs ± 1%  -15.80%  (p=0.000 n=10+10)
DictionaryPageBounds/DOUBLE        47.7µs ± 0%    39.8µs ± 1%  -16.57%  (p=0.000 n=6+10)
DictionaryPageBounds/BYTE_ARRAY    41.0µs ± 1%    36.1µs ± 1%  -11.86%  (p=0.000 n=10+9)
```
And deltas combining both optimizations:
```
name                             old time/op    new time/op    delta
DictionaryPageBounds/BOOLEAN       58.9µs ± 1%    21.9µs ± 0%  -62.93%  (p=0.000 n=10+9)
DictionaryPageBounds/INT32         65.6µs ± 0%    34.1µs ± 2%  -47.99%  (p=0.000 n=10+10)
DictionaryPageBounds/INT64         69.9µs ± 1%    37.6µs ± 1%  -46.13%  (p=0.000 n=9+10)
DictionaryPageBounds/INT96          127µs ± 1%      57µs ± 0%  -55.07%  (p=0.000 n=8+9)
DictionaryPageBounds/FLOAT         73.0µs ± 1%    38.7µs ± 1%  -46.97%  (p=0.000 n=10+10)
DictionaryPageBounds/DOUBLE        74.9µs ± 0%    39.8µs ± 1%  -46.92%  (p=0.000 n=10+10)
DictionaryPageBounds/BYTE_ARRAY    70.1µs ± 1%    36.1µs ± 1%  -48.45%  (p=0.000 n=10+9)
```

### DIFF
--- a/dictionary.go
+++ b/dictionary.go
@@ -28,10 +28,11 @@ type Dictionary interface {
 	Len() int
 
 	// Returns the dictionary value at the given index.
-	Index(int32) Value
+	Index(index int32) Value
 
-	// Inserts a value to the dictionary, returning the index at which the first
-	// value was recorded.
+	// Inserts a value to the dictionary and writes the indexes at which values
+	// were inserted. Returns the number of values inserted in the dictionary.
+	//Insert(indexes []int32, values []Value) int
 	Insert(Value) int32
 
 	// Given an array of dictionary indexes, lookup the values into the array
@@ -83,9 +84,8 @@ func (d *byteArrayDictionary) Index(i int32) Value {
 	return makeValueBytes(ByteArray, d.values.Index(int(i)))
 }
 
-func (d *byteArrayDictionary) Insert(v Value) int32 { return d.insert(v.ByteArray()) }
-
-func (d *byteArrayDictionary) insert(value []byte) int32 {
+func (d *byteArrayDictionary) Insert(v Value) int32 {
+	value := v.ByteArray()
 	if index, exists := d.index[string(value)]; exists {
 		return index
 	}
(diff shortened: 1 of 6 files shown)


dictionary.go
```diff
@@ -28,11 +28,11 @@ type Dictionary interface {
...
 
 	// Returns the dictionary value at the given index.
-	Index(int) Value
+	Index(index int32) Value
 
-	// Inserts a value to the dictionary, returning the index at which it was
-	// recorded.
-	Insert(Value) int
+	// Inserts a value to the dictionary and writes the indexes at which values
+	// were inserted.
```
kevinburkesegment, line 34: I think the doc could be clearer... do you mean to say values has the same length as indexes? How big of an array does the caller need to pass in, in other words

kevinburkesegment: Oh... I'd sort of reverse the order of these I guess, so the input is first and the  value that is being returned is second.

kevinburkesegment (approved):

dictionary.go
achille-roussel: I think Go tends to encourage having the output parameter first, considering the signatures of `copy`, `io.Copy`, or sending values to channels with `ch <- v`?

PR #108, opened 2022-03-18
fix github annotations

kevinburkesegment:
Previously Github would complain about the syntax when you ran the
    formatter using go 1.17, because the formatter doesn't know to ignore
    go 1.18 files.

    Instead only run the formatter on go 1.18 files.


kevinburkesegment:
Verified that this works by manually un-formatting a file and checking that the build failed.

achille-roussel (approved):
Nice!

PR #111, opened 2022-03-21
row_group_test: don't do work for test if we're not running it

kevinburkesegment:
Previously we'd always do the work in sortedRowGroup which is making
it difficult to troubleshoot the error reported in #89.

### DIFF
--- a/row_group_test.go
+++ b/row_group_test.go
@@ -106,210 +106,216 @@ func testSeekToRow(t *testing.T, newRowGroup func([]Person) parquet.RowGroup) {
 	}
 }
 
-var mergeRowGroupsTests = [...]struct {
+type mergeRowGroupTestCase struct {
 	scenario string
 	options  []parquet.RowGroupOption
 	input    []parquet.RowGroup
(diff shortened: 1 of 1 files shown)


row_group_test.go
```diff
@@ -106,210 +106,216 @@ func testSeekToRow(t *testing.T, newRowGroup func([]Person) parquet.RowGroup) {
...
-		options: []parquet.RowGroupOption{
-			parquet.SchemaOf(Person{}),
-		},
-		output: sortedRowGroup(
-			[]parquet.RowGroupOption{
+}
+
+// Use a function instead of an array because otherwise the sortedRowGroup()
+// calls will exercise the code even if you're not running these test cases.
+var mergeRowGroupsTests = func() []mergeRowGroupTestCase {
```
achille-roussel, line 118: Should this just be a function instead of a global variable?

achille-roussel (approved):

row_group_test.go
kevinburkesegment: yeah good call

PR #113, opened 2022-03-22
all: support repeated fixed length binary arrays

kevinburkesegment:
Previously when decoding a list of fixed length byte arrays, we'd pass
a byte array with capacity 1 to the decoder, which errored when
attempting to decode a 16 byte array - the capacity should be
a multiple of the length.

Fixes #89

### DIFF
--- a/dictionary.go
+++ b/dictionary.go
@@ -181,7 +181,7 @@ func newFixedLenByteArrayDictionary(typ Type, bufferSize int) *fixedLenByteArray
 	return &fixedLenByteArrayDictionary{
 		typ:    typ,
 		size:   size,
-		values: make([]byte, 0, dictCap(bufferSize, size)),
+		values: make([]byte, 0, dictCap(bufferSize, size)*size),
 	}
 }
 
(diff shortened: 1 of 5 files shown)


column_buffer.go
```diff
@@ -588,10 +588,11 @@ func (col *repeatedColumnBuffer) WriteRow(row Row) error {
 		return errRowHasTooFewValues(int64(len(row)))
 	}
 
-	col.rows = append(col.rows, region{
+	rgn := region{
```
achille-roussel, line 591: Style preference?

dictionary.go
```diff
@@ -349,7 +349,10 @@ func (page *indexedPage) RepetitionLevels() []int8 { return nil }
 
 func (page *indexedPage) DefinitionLevels() []int8 { return nil }
 
-func (page *indexedPage) WriteTo(e encoding.Encoder) error { return e.EncodeInt32(page.values) }
+func (page *indexedPage) WriteTo(e encoding.Encoder) error {
+	err := e.EncodeInt32(page.values)
+	return err
```
achille-roussel, line 354: Was this done by a formatting tool?

dictionary.go
```diff
@@ -181,7 +181,7 @@ func newFixedLenByteArrayDictionary(typ Type, bufferSize int) *fixedLenByteArray
 	return &fixedLenByteArrayDictionary{
 		typ:    typ,
 		size:   size,
-		values: make([]byte, 0, dictCap(bufferSize, size)),
+		values: make([]byte, 0, size),
```
achille-roussel, line 184: While it fixes the bug, it discard the buffer size hint; would this approach address the issue while retaining the use of `bufferSize`?
```suggestion
		values: make([]byte, 0, dictCap(bufferSize, size) * size),
```
`dictCap` returns a number of items, so if we scale it back to a number of bytes by multiplying by the size of an item it should address the issue?

dictionary.go
kevinburkesegment: No, another troubleshooting artifact (I just wanted to print the error)

achille-roussel (approved):
🚢

PR #115, opened 2022-03-24
Add new decimal primitive type

hhoughgg:
Added fixed_len_byte_array decimal option
Added timestamp and date to struct tags. 
Changed it so byte arrays are not always delta encoded if plain is set as a struct tag. Some readers such as spark vectorized reader do not support this encoding.

Would appreciate some feedback on this PR. Thanks

Also want to say thanks for this project! This performance is great

### DIFF
--- a/node.go
+++ b/node.go
@@ -504,7 +504,7 @@ func encodingAndCompressionOf(node Node) (encoding.Encoding, compress.Codec) {
 	// the opportunity to override this behavior if needed.
 	//
 	// https://github.com/apache/parquet-format/blob/master/Encodings.md#delta-length-byte-array-delta_length_byte_array--6
-	if node.Type().Kind() == ByteArray {
+	if node.Type().Kind() == ByteArray && len(node.Encoding()) == 0 {
 		encoding = &DeltaLengthByteArray
 	}
 
--- a/node.go
+++ b/node.go
@@ -496,6 +496,7 @@ func encodingAndCompressionOf(node Node) (encoding.Encoding, compress.Codec) {
 	// representations of the pages, picking the one with the smallest space
 	// footprint; keep it simple for now.
 	encoding := encoding.Encoding(&Plain)
+	nodeEncoding := node.Encoding()
 	compression := compress.Codec(&Uncompressed)
 	// The parquet-format documentation states that the
 	// DELTA_LENGTH_BYTE_ARRAY is always preferred to PLAIN when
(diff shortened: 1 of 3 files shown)


node.go
```diff
@@ -504,7 +504,7 @@ func encodingAndCompressionOf(node Node) (encoding.Encoding, compress.Codec) {
 	// the opportunity to override this behavior if needed.
 	//
 	// https://github.com/apache/parquet-format/blob/master/Encodings.md#delta-length-byte-array-delta_length_byte_array--6
-	if node.Type().Kind() == ByteArray {
+	if node.Type().Kind() == ByteArray && len(node.Encoding()) == 0 {
```
achille-roussel, line 507: `node.Encoding()` gets called twice in the function, do you think we could manage to just do a single call by capturing the result in a local variable?

schema.go
```diff
@@ -552,11 +557,28 @@ func makeStructField(f reflect.StructField) structField {
 					baseType = Int32Type
 				case reflect.Int64:
 					baseType = Int64Type
+				case reflect.Array:
+					length := int(math.Ceil((math.Log10(2) + float64(precision)) / math.Log10(256)))
```
achille-roussel, line 561: We're performing this same computation in two different places, what do you think about moving it to an internal function instead?

type.go
```diff
@@ -309,10 +310,11 @@ func (t *intType) ConvertedType() *deprecated.ConvertedType {
 //
 // https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#decimal
 func Decimal(scale, precision int, typ Type) Node {
-	switch typ {
-	case Int32Type, Int64Type:
+	length := int(math.Ceil((math.Log10(2) + float64(precision)) / math.Log10(256)))
+	switch typ.Kind() {
+	case Int32Type.Kind(), Int64Type.Kind(), FixedLenByteArrayType(length).Kind():
```
achille-roussel, line 315: This might be a simpler form of the same behavior?
```suggestion
	case Int32, Int64, FixedLenByteArray:
```

schema.go
hhoughgg: I went ahead and added a function for this but due to your other suggestion that simplifies the kind checking it is only being used once now. I still like the function suggestion since it allows adding some explanation. Let me know if any issues with the function naming or location.

type.go
```diff
@@ -304,15 +305,21 @@ func (t *intType) ConvertedType() *deprecated.ConvertedType {
 	return &convertedTypes[convertedType]
 }
 
+// FixedLenByteArray decimals are sized based on precision
+// this function calculates the necessary byte array size
+func CalcDecimalFixedLenByteArraySize(precision int) int {
```
achille-roussel, line 310: Would this function ever need to be called by users of this library? If not, I'd recommend we make it unexported, it's easier to export later if needed than take away and break programs.

achille-roussel (approved):
Thanks for the contribution!

PR #116, opened 2022-03-29
add parquet.Dictionary.Page method

achille-roussel:
Following up from user feedback, this PR adds the `parquet.Dictionary.Page` method, allowing the dictionary to be converted to a parquet page to read its content.

I first modified the `parquet.Dictionary` interface to add a `Values() ValueReader` method (in the same way we have on `parquet.Page`, and then realized that the interface and implementation started to overlap too much with pages, to the point that I estimated that it would be better to reduce the code surface to simply leverage the page APIs.

Prior to this change there was a symmetry between the `ReadFrom` and `WriteTo` methods of the `parquet.Dictionary` interface. By removing `WriteTo` (since it's now available by calling `Page().WriteTo`), this symmetry was broken so I replaced `ReadFrom` with a `parquet.Type.ReadDictionary` method instead. 

Please take a look and let me know if you have any comments!


column_buffer_go18.go
```diff
@@ -76,7 +76,8 @@ func (col *columnBuffer[T]) Write(b []byte) (int, error) {
 	if (len(b) % sizeof[T]()) != 0 {
 		return 0, fmt.Errorf("cannot write %s values from input of size %d", col.class.name, len(b))
 	}
-	return col.WriteRequired(cast.BytesToSlice[T](b))
+	n, err := col.WriteRequired(cast.BytesToSlice[T](b))
+	return sizeof[T]() * n, err
```
achille-roussel, line 80: This is a fix for a bug I encountered while working on the change.

dictionary.go
```diff
@@ -141,69 +164,88 @@ func (d *byteArrayDictionary) Bounds(indexes []int32) (min, max Value) {
...
+			copy(newValues, d.data)
+			d.data = newValues
+		}
+
+		n, err := decoder.DecodeFixedLenByteArray(d.size, d.data[len(d.data):cap(d.data)])
+		if n > 0 {
+			d.data = d.data[:len(d.data)+(n*d.size)]
+		}
+
+		if err != nil {
```
kevinburkesegment, line 218: This reads a little awkward... to mix the success and the error cases together in the same if statement, what about

```go
if err == io.EOF {
    return d, nil
}
if err != nil {
    return d, fmt.Errorf(...)
}
```

type_go18.go
```diff
@@ -248,12 +281,20 @@ func (t *timeType) NewColumnReader(columnIndex, bufferSize int) ColumnReader {
 	}
 }
 
+func (t *timeType) ReadDictionary(columnIndex, numValues int, decoder encoding.Decoder) (Dictionary, error) {
+	if t.Unit.Millis != nil {
```
kevinburkesegment, line 285: My hunch reading this would be that the `else` block - when millis is nil - is at a lower granularity than this side which then made it confusing why it needs to use more storage. 

Maybe you could use a method call? 

func (t timeType) IsUnix() bool 
func (t timeType) IsPrecise() bool 

or something? I don't know

kevinburkesegment: Not super important just was confusing reading it.

type_default.go
```diff
@@ -349,12 +382,28 @@ func (t *intType) NewColumnReader(columnIndex, bufferSize int) ColumnReader {
...
 }
 
+func (t *intType) ReadDictionary(columnIndex, numValues int, decoder encoding.Decoder) (Dictionary, error) {
+	if t.IsSigned {
+		if t.BitWidth == 64 {
+			return readInt64Dictionary(t, makeColumnIndex(columnIndex), numValues, decoder)
+		} else {
+			return readInt32Dictionary(t, makeColumnIndex(columnIndex), numValues, decoder)
+		}
+	} else {
```
kevinburkesegment, line 392: You can drop the else here because both sides of the if statement return.

type.go
```diff
@@ -155,6 +156,14 @@ type Type interface {
 	// parquet.Int32Reader interface to allow programs to more efficiently read
 	// columns of INT32 values.
 	NewColumnReader(columnIndex, bufferSize int) ColumnReader
+
+	// Reads a dictionary with values of this type from the decoder passed as
+	// argument.
+	//
+	// The number of values is a hint to optimize the allocation of memory
+	// buffers for the dictionary. Applications must pass zero for the number of
```
kevinburkesegment, line 164: "Callers that don't know how many values will be decoded should pass zero for numValues."

kevinburkesegment (approved):

type_go18.go
achille-roussel: There's two possible states here: either the millisecond or microsecond units are defined (it's a thrift union underneath), and the parquet format dictates that int32 is used for millisecond precision, int64 otherwise.

I'll add a method like you suggest to improve readability.

type_default.go
achille-roussel: True, but it loses symmetry in the method body, which I find to be a good way to convey intent as somewhat related conditions align on the same indentation levels.

achille-roussel:
Thanks for the useful comments!

PR #122, opened 2022-04-06
expose number of rows on parquet files

achille-roussel:
This PR adds a `NumRows() int64` method to the `parquet.File` type to offer a simple way to get the total number of rows in a parquet file.

### DIFF
--- a/file.go
+++ b/file.go
@@ -207,7 +207,7 @@ func (f *File) readPageIndex(section *bufferedSectionReader, decoder *thrift.Dec
 }
 
 // NumRows returns the number of rows in the file.
-func (f *File) NumRows() int64 { return f.metadata.numRows }
+func (f *File) NumRows() int64 { return f.metadata.NumRows }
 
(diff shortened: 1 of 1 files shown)


file.go
```diff
@@ -206,6 +206,9 @@ func (f *File) readPageIndex(section *bufferedSectionReader, decoder *thrift.Dec
 	return columnIndexes, offsetIndexes, nil
 }
 
+// NumRows returns the number of rows in the file.
+func (f *File) NumRows() int64 { return f.metadata.NumRows }
```
kevinburkesegment, line 210: Does this need to be on `*File` or should it just be on `File`?

kevinburkesegment (approved):

Pryz (approved):

file.go
achille-roussel: Either way would work, for the sake of consistency with every other methods I'd keep the receiver be a `*File`, making a copy when passing by value doesn't really add much to readability in my opinion.

PR #126, opened 2022-04-06
add parquet.ColumnMapping

achille-roussel:
This PR adds a new `parquet.ColumnMapping` type to the library, which intends to help applications perform conversions of column paths to their indexes in a row, which is useful when performing conversions of Go values into parquet rows for example.

### DIFF
--- a/column_mapping.go
+++ b/column_mapping.go
@@ -16,10 +16,12 @@ type ColumnMapping struct {
 }
 
 // ColumnIndex returns the column index of the column at the given path.
+// The path is the sequence of column names identifying a leaf column, starting
+// from the root column of the schema.
 //
 // If the path was not found in the mapping, or if it did not represent a
 // leaf column of the parquet schema, the method returns a negative value.
-func (m *ColumnMapping) ColumnIndex(path []string) (columnIndex int) {
+func (m *ColumnMapping) ColumnIndex(path ...string) (columnIndex int) {
 	return int(m.mapping.lookup(path))
 }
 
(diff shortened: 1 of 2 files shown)


column_mapping.go
```diff
@@ -0,0 +1,104 @@
...
+type ColumnMapping struct {
+	mapping columnMapping
+	columns [][]string
+}
+
+// ColumnIndex returns the column index of the column at the given path.
+//
+// If the path was not found in the mapping, or if it did not represent a
+// leaf column of the parquet schema, the method returns a negative value.
+func (m *ColumnMapping) ColumnIndex(path []string) (columnIndex int) {
```
kevinburkesegment, line 22: I'm a little confused about what would constitute a path in this example - normally I'd think of a path as a single string.

column_mapping.go
```diff
@@ -0,0 +1,104 @@
...
+// The method always returns the same slice value across calls to ColumnPaths,
+// applications should treat it as immutable.
+func (m *ColumnMapping) ColumnPaths() [][]string {
+	return m.columns
+}
+
+// String returns a string representation of the column mapping.
+func (m *ColumnMapping) String() string {
+	s := new(strings.Builder)
+	s.WriteString("{")
```
kevinburkesegment, line 37: WriteByte('{')?

kevinburkesegment (approved):

column_mapping.go
achille-roussel: It represents the sequence of column names that identify a leaf column in the parquet schema. We use `[]string` instead of a separator because all UTF-8 code points are valid in columns names, electing an arbitrary character as separator for a path representation would diverge from the specification and potentially conflict with column names found in parquet files that contained the separator.

PR #127, opened 2022-04-07
add parquet.Search

achille-roussel:
This PR adds a new `parquet.Search` function, which intends to help leverage column indexes to find the page that may contain a value. Since this is the main purpose of the column index, I think it makes sense to provide this feature in the library.

There are two cases:
* the index is in ascending order, in which case we can do a binary search on the min page values
* the index is not ordered, in which case we perform a linear scan for the first match

Technically we could potentially support another case to leverage descending ordering on column indexes, but I don't have any use cases for it at this time.

### DIFF
--- a/search.go
+++ b/search.go
@@ -36,7 +36,12 @@ func CompareNullsLast(cmp func(Value, Value) int) func(Value, Value) int {
 	}
 }
 
-// Search uses the column index passed as argument to find the page that the
+// Search is like Find, but uses the default ordering of the given type.
+func Search(index ColumnIndex, value Value, typ Type) int {
+	return Find(index, value, CompareNullsLast(typ.Compare))
+}
+
+// Find uses the column index passed as argument to find the page that the
 // given value is expected to be found in.
 //
 // The function returns the index of the first page that might contain the
@@ -45,8 +50,14 @@ func CompareNullsLast(cmp func(Value, Value) int) func(Value, Value) int {
 //
 // The comparison function passed as last argument is used to determine the
 // relative order of values. This should generally be the Compare method of
-// the column type.
-func Search(index ColumnIndex, value Value, cmp func(Value, Value) int) int {
+// the column type, but can sometimes be customized to modify how null values
+// are interpreted, for example:
+//
+//	pageIndex := parquet.Find(columnIndex, value,
+//		parquet.CompareNullsFirst(typ.Compare),
+//	)
+//
+func Find(index ColumnIndex, value Value, cmp func(Value, Value) int) int {
 	switch {
 	case index.IsAscending():
 		return binarySearch(index, value, cmp)
--- a/search.go
+++ b/search.go
@@ -1,5 +1,41 @@
 package parquet
 
+// CompareNullsFirst constructs a comparison function which assumes that null
+// values are smaller than all other values.
+func CompareNullsFirst(cmp func(Value, Value) int) func(Value, Value) int {
+	return func(a, b Value) int {
+		switch {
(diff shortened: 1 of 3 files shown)


kevinburkesegment (approved):

search.go
```diff
@@ -0,0 +1,66 @@
...
+// given value is expected to be found in.
+//
+// The function returns the index of the first page that might contain the
+// value. If the function determines that the value does not exist in the
+// index, NumPages is returned.
+//
+// The comparison function passed as last argument is used to determine the
+// relative order of values. This should generally be the Compare method of
+// the column type.
+func Search(index ColumnIndex, value Value, cmp func(Value, Value) int) int {
```
kevinburkesegment, line 13: Go 1.19 is going to add sort.Find with this type signature

```
+// Find uses binary search to find and return the smallest index i in [0, n)
+// at which cmp(i) <= 0. If there is no such index i, Find returns i = n.
+// The found result is true if i < n and cmp(i) == 0.
+// Find calls cmp(i) only for i in the range [0, n).
+//
+// To permit binary search, Find requires that cmp(i) > 0 for a leading
+// prefix of the range, cmp(i) == 0 in the middle, and cmp(i) < 0 for
+// the final suffix of the range. (Each subrange could be empty.)
+// The usual way to establish this condition is to interpret cmp(i)
+// as a comparison of a desired target value t against entry i in an
+// underlying indexed data structure x, returning <0, 0, and >0
+// when t < x[i], t == x[i], and t > x[i], respectively.
+//
+// For example, to look for a particular string in a sorted, random-access
+// list of strings:
+//    i, found := sort.Find(x.Len(), func(i int) int {
+//        return strings.Compare(target, x.At(i))
+//    })
+//    if found {
+//        fmt.Printf("found %s at entry %d\n", target, i)
+//    } else {
+//        fmt.Printf("%s not found, would insert at %d", target, i)
+//    }
+func Find(n int, cmp func(int) int) (i int, found bool) {
```

Should we try to match that?

peterdemartini, line 13: I am not intimately familiar with parquet but I think it might be possible that more than one page can contain the same value. If so, should this return multiple page indices?

Additionally, will the name of this function cause conflicts with other types of searches, like searching within a page to find matches of row indices or values?

achille-roussel: Sounds good 👍

search_test.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	formatIndex := indexer.ColumnIndex()
+	columnIndex := parquet.NewColumnIndex(parquet.Int32, &formatIndex)
+
+	for i, values := range pages {
+		t.Run(fmt.Sprintf("page#%02d", i), func(t *testing.T) {
+			for _, value := range values {
+				v := parquet.ValueOf(value)
+				j := parquet.Search(columnIndex, v, parquet.Int32Type.Compare)
```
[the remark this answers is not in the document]
achille-roussel: There is a test below with MinInt32 and MaxInt32 for this, let me know if you have suggestions for other values to test.

search.go
achille-roussel: The intent would be for the application to start reading pages from the first and forward, assuming values are ordered, it can stop of the first page that doesn't contain the value it's looking for. If values are not ordered, it needs to scan everything either ways, so no real benefits of returning a list of all page indexes in that case.

search_test.go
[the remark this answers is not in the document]
kevinburkesegment: Oh ok thumbsup

PR #129, opened 2022-04-07
allow configuring sorting columns on parquet writers

achille-roussel:
This PR adds the ability to configure sorting columns on parquet writers. This is useful when applications don't have an instance of `parquet.RowGroup` to write and instead write individual rows to the writer.

I made the `parquet.SortingColumns` option constructor return a value which implements both `parquet.RowGroupOption` and `parquet.WriterOption`, which I have mixed feelings about. I was considering also exporting a type like this:
```go
type SortingColumns []SortingColumn
```
However, it creates an asymmetry with option constructors like `parquet.BloomFilters`, maybe making the code harder to discover? 🤷 

Happy to get feedback, for now I'm just taking the path of least resistance with this change.

### DIFF
--- a/config.go
+++ b/config.go
@@ -139,6 +139,7 @@ type WriterConfig struct {
 	DataPageStatistics   bool
 	KeyValueMetadata     map[string]string
 	Schema               *Schema
+	SortingColumns       []SortingColumn
 	BloomFilters         []BloomFilterColumn
 }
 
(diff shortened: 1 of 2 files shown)


config.go
```diff
@@ -404,13 +406,26 @@ func ColumnBufferSize(size int) RowGroupOption {
 // The order of sorting columns passed as argument defines the ordering
 // hierarchy; when elements are equal in the first column, the second column is
 // used to order rows, etc...
-func SortingColumns(sortingColumns ...SortingColumn) RowGroupOption {
+func SortingColumns(columns ...SortingColumn) interface {
```
kevinburkesegment, line 409: Why wouldn't this just return []SortingColumn?

kevinburkesegment: Ohhh... the RowGroupOption is part of the return interface, that's not super intuitive. I'd maybe pull that out, `type RowGroupWriterOption` (?)

kevinburkesegment (approved):
I don't have the context to say really about the wider interface :( I trust your judgement

config.go
achille-roussel: The return type needs to implement the option interfaces to be used as input argument to `parquet.NewWriter` and such.

PR #130, opened 2022-04-08
fix file buffer pool

achille-roussel:
While working on using the file buffer pool from this package, I ran into an issue caused by the fact that `os.File` has a single cursor for read AND write operations, which resulted in programs being unable to read back the data that they had written to the buffers.

This PR addresses the issue by adding a second cursor to track the position of read operations, mimicking the behavior of `bytes.Buffer` on top of disk storage.

I also added tests, documentation, and fixed a few typos.


README.md
```diff
@@ -496,6 +496,41 @@ achieved by using an intermediary buffering layer with `parquet.Buffer`.
...
 for the full interface documentation.
 
+### C. Using on-disk page buffers
+
+When generating parquet files, the writer needs to buffer all pages before it
+can create the row group. This may require significant amounts of memory as the
+entire file content must be buffered prior to generating it. In some cases, the
+files might even be larger than the amount of memory available to the program.
+
+The `paruqet.Writer` can be configured to use disk storage instead as a scratch
```
Pryz, line 506:
```suggestion
The `parquet.Writer` can be configured to use disk storage instead as a scratch
```

Pryz (approved):

README.md
achille-roussel: Thanks!

PR #133, opened 2022-04-12
Expose concatenation of row groups as MultiRowGroup

brancz:
Happy to discuss the API more, but wanted to open this PR to get that conversation started.


brancz:
I need concat for the same reason as the `parquet.Reader` does, to concatenate file pages that have the identical schema. What was nice about that so far was that there wasn't a complicated schema comparison involved. Maybe we can find a fast path for comparing schemas of row groups? Then I think I like the `MultiRowGroup` the best.

If I'm reading this correctly, I think that `fileRowGroup`s of the same `parquet.File` will have the identical struct pointer, so maybe we can do a pointer comparison as the fast path?

https://github.com/segmentio/parquet-go/blob/c2916cf64256c9be5f3e69817938c9a9f5893d95/file.go#L88-L95

brancz:
Updated with the discussed API and fast path. Let me know what you think! I was contemplating making rowgroups variadic, but then we would have a harsher breaking change later if we wanted to add functional options as the other APIs have (I could also add them now, but don't really have a use yet which is why I left it out for now).

multirowgroup.go
```diff
@@ -0,0 +1,246 @@
...
+
+import (
+	"io"
+)
+
+// MultiRowGroup wraps multiple row groups to appear as if it was a single
+// RowGroup. RowGroups must have the same schema or it will error.
+func MultiRowGroup(rowGroups []RowGroup) (RowGroup, error) {
+	if len(rowGroups) == 0 {
+		return nil, nil
```
achille-roussel, line 11: This should return a non-nil value, otherwise we are creating fertile ground for application to trigger panics dereferencing nil pointers:
```suggestion
		return emptyRowGroup{}, nil
```

multirowgroup.go
```diff
@@ -0,0 +1,246 @@
+package parquet
+
+import (
+	"io"
+)
+
+// MultiRowGroup wraps multiple row groups to appear as if it was a single
+// RowGroup. RowGroups must have the same schema or it will error.
+func MultiRowGroup(rowGroups []RowGroup) (RowGroup, error) {
```
achille-roussel, line 9: To match the model of `io.MultiReader`, how about using an argument list instead?
```suggestion
func MultiRowGroup(rowGroups ...RowGroup) (RowGroup, error) {
```
I would also prefer not to return an error here and have the signature be
```suggestion
func MultiRowGroup(rowGroups ...RowGroup) RowGroup
```
Panicking on invalid input seems fine to me as it's an invalid use of the function.

multirowgroup.go
```diff
@@ -0,0 +1,246 @@
+package parquet
```
achille-roussel, line 1: To follow the naming schema of other files, this one should probably be renamed to `multi_row_group.go`.

Alternatively, the code could also be homed in `row_group.go`, up to you!

multirowgroup.go
```diff
@@ -0,0 +1,246 @@
...
+
+		for j, rowGroup := range rowGroups {
+			c.columns[i].chunks[j] = rowGroup.Column(i)
+		}
+	}
+
+	return nil
+}
+
+func compatibleSchema(rowGroups []RowGroup) (*Schema, error) {
```
achille-roussel, line 45: To follow the naming scheme of functions with similar purpose:
```suggestion
func compatibleSchemaOf(rowGroups []RowGroup) (*Schema, error) {
```

multirowgroup.go
```diff
@@ -0,0 +1,246 @@
...
+	c := new(multiRowGroup)
+	c.init(rowGroups)
+	return c, nil
+}
+
+func (c *multiRowGroup) init(rowGroups []RowGroup) error {
+	schema, err := compatibleSchema(rowGroups)
+	if err != nil {
+		return err
+	}
```
achille-roussel, line 26: How about moving this to `MultiRowGroup` and accepting the schema as argument? That way we can bypass the re-validation when calling this method internally in places where `concatenedRowGroup` was used before.
```suggestion
func (c *multiRowGroup) init(schema *Schema, rowGroups []RowGroup) error {
```

achille-roussel (approved):

PR #134, opened 2022-04-12
Support column indexes over multiple pages

lmarburger:
This PR started when noticing a small bug in the column index implementation and exploring how to exercise it in a test case. That resulted in noticing that some behavior wasn't covered by tests because the tests only created single page columns.

The big change is to update the tests to create multipage columns. That exposes a few small bugs in the implementation which are fixed in subsequent commits.

The fundamental change in behavior that this PR introduces is that `internal/bits/order` should return `2` when the input slice contains all equal values. This is necessary so that a column index can report an accurate order when one of the sets of values is ordered and the other is all equal value. For example, a column index with the metadata `min=[0,0,0] max=[6,5,4]` should report descending. I think this is likely to come up when a column can contain null values. The min index will probably be a set of all nulls so it should defer to the ordering of the max index.

I highly suggest reviewing this PR a commit at a time. Each commit is a focused change with a commit message describing it. Various assembly implementations will need to be updated. I'm hoping that CI will expose those paths and expose them as failures.

### DIFF
--- a/column_index.go
+++ b/column_index.go
@@ -145,7 +145,7 @@ func (i byteArrayColumnIndex) NullCount(int) int64 { return 0 }
 func (i byteArrayColumnIndex) NullPage(int) bool   { return false }
 func (i byteArrayColumnIndex) MinValue(int) Value  { return makeValueBytes(ByteArray, i.page.min()) }
 func (i byteArrayColumnIndex) MaxValue(int) Value  { return makeValueBytes(ByteArray, i.page.max()) }
-func (i byteArrayColumnIndex) IsAscending() bool   { return bytes.Compare(i.page.bounds()) < 0 }
+func (i byteArrayColumnIndex) IsAscending() bool   { return bytes.Compare(i.page.bounds()) <= 0 }
 func (i byteArrayColumnIndex) IsDescending() bool  { return bytes.Compare(i.page.bounds()) > 0 }
 
 type fixedLenByteArrayColumnIndex struct{ page *fixedLenByteArrayPage }
@@ -159,7 +159,7 @@ func (i fixedLenByteArrayColumnIndex) MinValue(int) Value {
 func (i fixedLenByteArrayColumnIndex) MaxValue(int) Value {
 	return makeValueBytes(FixedLenByteArray, i.page.max())
 }
-func (i fixedLenByteArrayColumnIndex) IsAscending() bool  { return bytes.Compare(i.page.bounds()) < 0 }
+func (i fixedLenByteArrayColumnIndex) IsAscending() bool  { return bytes.Compare(i.page.bounds()) <= 0 }
 func (i fixedLenByteArrayColumnIndex) IsDescending() bool { return bytes.Compare(i.page.bounds()) > 0 }
 
 // The ColumnIndexer interface is implemented by types that support generating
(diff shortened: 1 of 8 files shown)


column_index.go
```diff
@@ -364,6 +364,27 @@ func splitFixedLenByteArrayList(size int, data []byte) [][]byte {
 }
 
 func boundaryOrderOf(minOrder, maxOrder int) format.BoundaryOrder {
+	if minOrder == 2 || maxOrder == 2 {
```
kevinburkesegment, line 367: I've been reading through code for 5 minutes now to try to figure out what "2" means in this context... can we declare a variable with value 2, and then refer to the variable in the function? Or add a comment to this function, or to internal/bits/order.go, or ?

column_test.go
```diff
@@ -397,3 +410,114 @@ func (i *fileOffsetIndex) CompressedPageSize(j int) int64 {
...
+	}
+}
+
+func (c *columnStats) pageRead() {
+	c.page++
+}
+
+type indexOrder int
+
+const (
```
kevinburkesegment, line 445: Ah, is it this??

lmarburger: Basically, yes. This is a const representation of the expected values.

column_index.go
lmarburger: Currently, `internal/bits/order` returns `-1, 0, 1` to mean [ordered descending, unordered, and ordered ascending](https://github.com/segmentio/parquet-go/blob/54f6b56fa4371c350c43fbb91c648993f9053157/internal/bits/order_amd64.s#L5-L7). This PR adds `2` to mean `equal`.

Ideally, these would all be constants. The one factor complicating those efforts is the assembly. https://github.com/segmentio/parquet-go/blob/54f6b56fa4371c350c43fbb91c648993f9053157/internal/bits/order_amd64.s

kevinburkesegment (approved):

achille-roussel (commented):
It seems we are binding a few fixes into a single change, do you think we could try to split the efforts into a couple of PRs?

I'm a bit concerned about my typo of `[0]` instead of `[i]` turning into having to update vectorized assembly algorithms, it seems we could take a more incremental path here.

We might also want to think about the complexity/benefit ratio; ideally we handle all edge cases, but if they occur so rarely in practice, are they worth the cost? Defining the ordering of column indexes is an optimization, worst case we do linear scans over the index but if it occurs in a sub-decimal fraction of the queries, all we'll see is higher latency on the p99+, the output would remain correct.

Let me know what you think!

achille-roussel:
> The min index will probably be a set of all nulls so it should defer to the ordering of the max index.

From my understanding of the parquet format's spec, the column index will not contain nulls in min/max values, which is why the column index has a boolean array to determine whether a page contains only nulls (in which case there are no min/max).

I believe this documentation would be the reference here https://github.com/apache/parquet-format/blob/master/src/main/thrift/parquet.thrift#L939-L946

lmarburger:
I think I understand what you're saying. What tests could we write that would pass the order implementation as it exists today without an equality state? In explaining this, I created #138 which is this the core of this PR. It expands test coverage and fixes the one bug in string slice ordering.

lmarburger:
Closing this for now as the core of this behavior has been merged in #138

PR #137, opened 2022-04-13
Fix column index null count

lmarburger:
Fixes a simple typo in `NullCount()`. There is no test change in this PR because no tests exercise this code path. Expanding test coverage is a larger change started in #134


achille-roussel (approved):
🚢

PR #138, opened 2022-04-13
Core multi-page column index support

lmarburger:
This PR was extracted from #134. It pulls the multi-page tests without the need for adding an equality state. As such, it focuses only on expanding test coverage, achieving order parity between buffer and file indexes, and fixing a bug in ordering string slices.

### DIFF
--- a/column_index.go
+++ b/column_index.go
@@ -1,8 +1,6 @@
 package parquet
 
 import (
-	"bytes"
-
 	"github.com/segmentio/parquet-go/encoding"
 	"github.com/segmentio/parquet-go/format"
 	"github.com/segmentio/parquet-go/internal/bits"
@@ -145,8 +143,8 @@ func (i byteArrayColumnIndex) NullCount(int) int64 { return 0 }
 func (i byteArrayColumnIndex) NullPage(int) bool   { return false }
 func (i byteArrayColumnIndex) MinValue(int) Value  { return makeValueBytes(ByteArray, i.page.min()) }
 func (i byteArrayColumnIndex) MaxValue(int) Value  { return makeValueBytes(ByteArray, i.page.max()) }
-func (i byteArrayColumnIndex) IsAscending() bool   { return bytes.Compare(i.page.bounds()) < 0 }
-func (i byteArrayColumnIndex) IsDescending() bool  { return bytes.Compare(i.page.bounds()) > 0 }
+func (i byteArrayColumnIndex) IsAscending() bool   { return false }
+func (i byteArrayColumnIndex) IsDescending() bool  { return false }
 
 type fixedLenByteArrayColumnIndex struct{ page *fixedLenByteArrayPage }
 
@@ -159,8 +157,8 @@ func (i fixedLenByteArrayColumnIndex) MinValue(int) Value {
 func (i fixedLenByteArrayColumnIndex) MaxValue(int) Value {
 	return makeValueBytes(FixedLenByteArray, i.page.max())
 }
-func (i fixedLenByteArrayColumnIndex) IsAscending() bool  { return bytes.Compare(i.page.bounds()) < 0 }
-func (i fixedLenByteArrayColumnIndex) IsDescending() bool { return bytes.Compare(i.page.bounds()) > 0 }
+func (i fixedLenByteArrayColumnIndex) IsAscending() bool  { return false }
+func (i fixedLenByteArrayColumnIndex) IsDescending() bool { return false }
 
 // The ColumnIndexer interface is implemented by types that support generating
 // parquet column indexes.
--- a/column_test.go
+++ b/column_test.go
@@ -144,6 +144,7 @@ func checkColumnChunkColumnIndex(columnChunk parquet.ColumnChunk) error {
 	columnIndex := columnChunk.ColumnIndex()
 	numPages := columnIndex.NumPages()
 	pagesRead := 0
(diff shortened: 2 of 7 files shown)


internal/bits/order_test.go
```diff
@@ -349,7 +349,8 @@ func TestOrderOfBytes(t *testing.T) {
 	err := quickCheck(func(values [][16]byte) bool {
 		slices := make([][]byte, len(values))
 		for i, v := range values {
-			slices[i] = v[:]
+			cpy := v
+			slices[i] = cpy[:]
```
achille-roussel, line 353: Why did we need the copy here?

column_index.go
```diff
@@ -145,8 +143,8 @@ func (i byteArrayColumnIndex) NullCount(int) int64 { return 0 }
 func (i byteArrayColumnIndex) NullPage(int) bool   { return false }
 func (i byteArrayColumnIndex) MinValue(int) Value  { return makeValueBytes(ByteArray, i.page.min()) }
 func (i byteArrayColumnIndex) MaxValue(int) Value  { return makeValueBytes(ByteArray, i.page.max()) }
-func (i byteArrayColumnIndex) IsAscending() bool   { return bytes.Compare(i.page.bounds()) < 0 }
-func (i byteArrayColumnIndex) IsDescending() bool  { return bytes.Compare(i.page.bounds()) > 0 }
+func (i byteArrayColumnIndex) IsAscending() bool   { return true }
```
achille-roussel, line 146: Is the reasoning that it is safe for this to be in ascending order because there is only one page?

I'm also wondering whether it would be the right decision to make, maybe we should assume that we don't know what the order is if there is a single page?

internal/bits/order_test.go
lmarburger: This was causing each item in `slices` to be a reference to the last item in `values`. Assigning it to an intermediate var was one way to resolve this.

column_index.go
lmarburger: That's correct. We basically have to pick one and my take is that the choice has already been made for us by `fileColumnIndex`. It uses `internal/bits/order` to determine order which [returns ascending for a single-length slice](https://github.com/segmentio/parquet-go/blob/9409f0293efe3f41863e20a01a07d076bfe6950e/internal/bits/order.go#L56-L57). It has a vectorized assembly implementation which is harder to change than this code here.

achille-roussel: The `bits.OrderOf*` functions are implemented in Go right? only the internal `bits.orderOf*` functions have assembly-optimized version, or am I reading this wrong?

internal/bits/order_test.go
achille-roussel: Ah interesting, the range variables are always full of surprises.

How about this version so we don't need the copy?
```go
for i := range values {
    slices[i] = values[i][:]
}
```

column_index.go
lmarburger: 🤦 You're right. We could change the implementation of the `len == 1` case globally without touching the optimized code.

I'll update the code to treat single-item slices as unordered.

internal/bits/order_test.go
lmarburger: Changed in f55e598

column_index.go
lmarburger: Updated in db705aa

column_test.go
```diff
@@ -397,3 +410,110 @@ func (i *fileOffsetIndex) CompressedPageSize(j int) int64 {
...
+			order = next
+			continue
+		}
+		if order != next {
+			return unorderedIndexOrder
+		}
+	}
+
+	switch order {
+	case -1:
```
achille-roussel, line 514: The `parquet.Type.Compare` method promises that the values will be negative, zero, or positive, but not equal to -1, 0, or +1 https://pkg.go.dev/github.com/segmentio/parquet-go#Type

How about using `< 0` instead here?

`parquet.Type.Compare` will return a negative value if the first value is less than the second, in which case the order is ascending as well isn't it? It seems this logic is inverted then?

internal/bits/order_test.go
```diff
@@ -86,7 +86,7 @@ func isUndefined(ordering int) bool {
 }
 
 func isSorted(set sort.Interface) bool {
```
achille-roussel, line 88: How about renaming this to `isOrdered`? I feel like having it named `isSorted` but having a different behavior than `stort.IsSorted` is a bit confusing.

column_test.go
lmarburger: > `parquet.Type.Compare` will return a negative value if the first value is less than the second, in which case the order is ascending as well isn't it? It seems this logic is inverted then?

That's correct. This code is calling [`Compare(second, first)`](https://github.com/segmentio/parquet-go/blob/1b616d0ef40d6ac4d55c1448c582dd4edd6980d7/column_test.go#L500) because I think in a previous version I was comparing it to `OrderBy*()`. That's confusing so I updated the code to do what's expected.

Fixed that and better adherence to the `parquet.Type.Compare` API in 1aa9ce4

internal/bits/order_test.go
lmarburger: 👍 c229acb

column_test.go
achille-roussel: Thanks for the clarification, it was confusing indeed to have `Compare(second, first)` 👍

achille-roussel (approved):
🙌

PR #140, opened 2022-04-15
support unordered columns

### LINKED ISSUE #119: How to keep the schema field same order as struct?
Hello,

I have a table already in redshift, I need to keep the same schema (even the order of the fields) to load parquet to redshift.

However I found the library creates the schema based on the struct and sorts the key. what I need is to keep the schema field same order as we defined in struct.

In redshift I have:
```
Schema: message test {
  required binary b (STRING);
  required int32 a (INT(32, true));
  required binary c (STRING);
}
```

The defined struct:
```
	type Test struct {
		B string `parquet:"b,plain,gzip"`
		A string `parquet:"a,plain,gzip"`
		C string `parquet:"c,plain,gzip"`
	}

```

its generated schema:
```
Schema: message Test {
  required binary a (STRING);
  required binary b (STRING);
  required binary c (STRING);
}

```

Is there any way to do that?

achille-roussel:
Fixes #119 

This PR contains a refactor of the package to respect column orders instead of forcing it to be sorted.

As discussed in #119, I initially attempted to add an option to support configuring the layout of columns in parquet files, which appeared to be a smaller scope change. However, it uncovered a bug where reading parquet files with unsorted column names would break due to the package lacking mapping between the on-file and in-memory column orders (see https://github.com/segmentio/parquet-go/commit/0912c251601f5cc0f2df3544df83a735443ff9f2).

I resorted in making a change to relax the requirement to have columns ordered by names in the parquet schema, respecting the order in which the columns are declared, when constructing schemas from Go values or reading them from parquet files.

To achieve the desired behavior while maintaining simple APIs, I made breaking changes to the `parquet.Node` interface:
* introduced a `parquet.Field` type to represent node fields
* introduced `Leaf() bool` and `Fields() []Field` methods on `parquet.Node`
* removed the `NumChildren`, `ChildNames`, `ChildByName`, and `ValueByName` methods on `parquet.Node`
* removed the `parquet.IndexedNode` interface

While most translations were straightforward, the code in `covert.go` was strongly coupled to the assumption that columns would be sorted by name, and had to be rewritten to account for the relaxed constraints.

A few tests that also relied on column ordering had to be adapted.

Overall, while this is a significant change, I believe that it makes the package more robust and more useful: it fixes the problems of not being able to read parquet with unordered schemas by removing all assumptions of column ordering, and brings meaningful improvements to exported APIs like `parquet.Node`, all of which seem like worthy additions to me.

Let me know if you have any questions or feedback.


node.go
```diff
@@ -38,29 +38,18 @@ type Node interface {
...
 	//
-	// The method panics if it is called on a leaf node.
-	ChildByName(name string) Node
-
-	// ValueByName is returns the sub-value with the given name in base.
-	ValueByName(base reflect.Value, name string) reflect.Value
+	// This method returns an empty mapping when called on leaf nodes.
+	Fields() []Field
 
-	// Returns the list of encodings used by the node and its children.
```
achille-roussel, line 63: I changed this requirement since it was the source of significant code complexity but was not useful: only leaf columns are encoded in parquet, there is no point exposing encodings for non-leaf columns.

node.go
```diff
@@ -69,7 +58,7 @@ type Node interface {
 	// this method. Applications should treat the return value as immutable.
 	Encoding() []encoding.Encoding
 
-	// Returns the list of compression codecs used by the node and its children.
```
achille-roussel, line 72: Same reasoning of for `Encoding() []encoding.Encoding` here.

column_mapping.go
```diff
@@ -40,7 +41,8 @@ func (m *ColumnMapping) String() string {
 
 	if len(m.columns) > 0 {
 		for _, path := range m.columns {
-			fmt.Fprintf(s, "\n  % 2d => %q", m.ColumnIndex(path...), columnPath(path))
+			columnIndex, _ := m.Lookup(path...)
+			fmt.Fprintf(s, "\n  % 2d => %q", columnIndex, columnPath(path))
```
kevinburkesegment, line 45: Just a guess that most pages will have 99 columns or fewer?

achille-roussel: I think I was more concerned that the visual impact will be felt when displaying 10-20 columns, beyond hundreds this format is likely the wrong tool and if the alignment is inaccurate it's not the worst part of the experience anyways.

merge.go
```diff
@@ -234,6 +234,7 @@ func (cur *bufferedRowGroupCursor) nextRowValuesOf(values []Value, columnIndex i
 	return append(values, cur.columns[columnIndex]...)
 }
 
+/*
```
kevinburkesegment, line 237: ?

node.go
```diff
@@ -306,6 +270,17 @@ func fieldRepetitionTypeOf(node Node) *format.FieldRepetitionType {
 	}
 }
 
+func fieldRepetitionTypeOf(node Node) format.FieldRepetitionType {
+	switch {
+	case node.Optional():
+		return format.Optional
+	case node.Repeated():
+		return format.Repeated
```
kevinburkesegment, line 278: Maybe more future proof:

```
case node.Required():
    return format.Required
default:
    panic("unknown node repetition type")
```

although maybe it's hard to think of what a fourth or fifth type would be.

kevinburkesegment (approved):

node.go
achille-roussel: A risk is that the state may come from user input (say an exotic parquet file, or custom implementation of the Node interface), panic may not be the desired behavior then, but maybe validation should happen at a different stage as well... it seems hard to perceive what the right trade off would be.

PR #141, opened 2022-04-16
fix issue 121

### LINKED ISSUE #121: Incorrect data written and panic for nested repeated columns when using parquet.Buffer
I believe the logic in [repeatedPage.Slice](https://github.com/segmentio/parquet-go/blob/main/page.go#L446) is not correctly handling when a repeated column is nested by multiple levels.  Two issues are presented and there are minimal failing tests committed [here](https://github.com/mdisibio/parquet-go/blob/buffer-nested-test/buffer_test.go#L426).  The tests attempt to roundtrip some basic objects.

1.  For a normal column, it writes incorrect data and the roundtrip detects mismatch.
2. For an optional column with only nulls, negative offsets `i` and `j` are calculated in `repeatedPage.Slice` which leads to slice panic.

An observation:
* It occurs when enough data is written to exceed the first page and a split is determined in `forEachPageSlice`.  
* The issues occur in the calculation of subsequent pages.  First page is always fine.
* The tests aim for minimal reproduction by using page size of 100 bytes and only write a few objects.

Test 1:
The expected objects are [i, i] ([0,0] ... [5,5]).  The 6 objects are written as 2 pages of 3.  The mismatch starts on the 2nd page.
```go
func TestBufferRoundtripNestedRepeated(t *testing.T) {
	type C struct {
		D int
	}
	type B struct {
		C []C
	}
	type A struct {
		B []B
	}

	// Write enough objects to exceed first page
	buffer := parquet.NewBuffer()
	var objs []A
	for i := 0; i < 6; i++ {
		o := A{[]B{{[]C{
			{i},
			{i},
		}}}}
		buffer.Write(&o)
		objs = append(objs, o)
	}

	buf := new(bytes.Buffer)
	w := parquet.NewWriter(buf, parquet.PageBufferSize(100))
	w.WriteRowGroup(buffer)
	w.Flush()
	w.Close()

	r := parquet.NewReader(bytes.NewReader(buf.Bytes()))
	for i := 0; ; i++ {
		o := new(A)
		err := r.Read(o)
		if err == io.EOF {
			break
		}
		if !reflect.DeepEqual(*o, objs[i]) {
			t.Errorf("points mismatch at row index %d: want=%v got=%v", i, objs[i], o)
		}
	}
}
```

Test output:
```
$ go test -v -run=TestBufferRoundtripNestedRepeated
=== RUN   TestBufferRoundtripNestedRepeated
    buffer_test.go:463: points mismatch at row index 3: want={[{[{3} {3}]}]} got=&{[{[{1} {2}]}]}
    buffer_test.go:463: points mismatch at row index 4: want={[{[{4} {4}]}]} got=&{[{[{2} {3}]}]}
    buffer_test.go:463: points mismatch at row index 5: want={[{[{5} {5}]}]} got=&{[{[{3} {4}]}]}
--- FAIL: TestBufferRoundtripNestedRepeated (0.00s)
FAIL
exit status 1
FAIL	github.com/segmentio/parquet-go	0.013s
```

Test2:
Virtually similar except `*int` data type and writes 26 objects of nils to exceed first page.

```
$ go test -v -run=TestBufferRoundtripNestedRepeatedOptional
=== RUN   TestBufferRoundtripNestedRepeatedOptional
--- FAIL: TestBufferRoundtripNestedRepeatedOptional (0.00s)
panic: runtime error: slice bounds out of range [:-13] [recovered]
	panic: runtime error: slice bounds out of range [:-13]

goroutine 19 [running]:
testing.tRunner.func1.2({0x140f1e0, 0xc00015e090})
	/usr/local/go/src/testing/testing.go:1209 +0x24e
testing.tRunner.func1()
	/usr/local/go/src/testing/testing.go:1212 +0x218
panic({0x140f1e0, 0xc00015e090})
	/usr/local/go/src/runtime/panic.go:1038 +0x215
github.com/segmentio/parquet-go.(*int64Page).Slice(0xc00034001a, 0x0, 0x0)
	/Users/marty/src/oss/parquet-go/page_default.go:262 +0xd3
github.com/segmentio/parquet-go.(*repeatedPage).Slice(0xc00012c230, 0xd, 0x1a)
	/Users/marty/src/oss/parquet-go/page.go:494 +0x19f
github.com/segmentio/parquet-go.forEachPageSlice({0x14d2ab8, 0xc00012c230}, 0x64, 0xc0001a3ba8)
	/Users/marty/src/oss/parquet-go/page.go:202 +0x103
```

achille-roussel:
Fixes #121 

This PR addresses the bug reported in issue 121, which were actually a combination of two problems:
* when slicing a page of repeated values, the math to compute the indexes to slice in the underlying page was done wrong
* when reading values from a repeated page containing nulls, the column index was not properly set on the returned values

As part of the investigation, I added new functions to print parquet data structures, which were useful to introspect the program state in the tests that highlighted the issue, and come up with a fix.


go.mod
```diff
@@ -11,3 +11,8 @@ require (
 	github.com/segmentio/encoding v0.3.3
 	golang.org/x/sys v0.0.0-20211110154304-99a53858aa08
 )
+
+require (
+	github.com/mattn/go-runewidth v0.0.9 // indirect
+	github.com/olekukonko/tablewriter v0.0.5 // indirect
```
kevinburkesegment, line 17: Surprised these are listed as indirect dependencies, we use them directly in print.go.

print.go
```diff
@@ -177,6 +192,142 @@ var (
...
+	for {
+		if row, err = rows.ReadRow(row[:0]); err != nil {
+			if !errors.Is(err, io.EOF) {
+				return err
+			}
+			break
+		}
+
+		cells = cells[:0]
+		fmt.Printf("%#v\n", row)
```
kevinburkesegment, line 240: Stray, I think?

kevinburkesegment (approved):

print.go
achille-roussel: Yes, that was a miss, removed in https://github.com/segmentio/parquet-go/pull/146

go.mod
achille-roussel: I think I did the `go get` before writing the `import` statement in the code, `go mod tidy` cleaned that up 👍

Issue #124, opened 2022-04-06
Add the ability to configure the compression for all Parquet fields


Pryz:
Follow up from https://github.com/segmentio/parquet-go/issues/114

To simplify the usage of compression, we want to be able to configure the compression algorithm to be used against all the fields when creating a writer.

achille-roussel:
Fixed in #143

PR #143, opened 2022-04-16
support configuring default compression codec

achille-roussel:
Based on #140, this PR addresses #124 by adding a new writer option to configure the default compression codec to use when none are specified on the schema.

As part of this work, I am suggesting to simplify further the `parquet.Node` interface to support having a single encoding and compression codec per node. Prior to this change, it was possible to configure multiple encodings and compression codecs on a column; however, we never made use of this feature, nor did it provide applications with information they couldn't other access more precisely with different mechanism (see the comment I left in `column.go` for details).

### DIFF
--- a/column.go
+++ b/column.go
@@ -306,7 +306,7 @@ func (cl *columnLoader) open(file *File, path []string) (*Column, error) {
 		if len(c.chunks) > 0 {
 			// Pick the encoding and compression codec of the first chunk.
 			//
-			// Techically each column chunk may use a different compression
+			// Technically each column chunk may use a different compression
 			// codec, and each page of the columm chunk might have a different
 			// encoding. Exposing these details does not provide a lot of value
 			// to the end user.
--- a/column.go
+++ b/column.go
@@ -27,8 +27,8 @@ type Column struct {
 	chunks      []*format.ColumnChunk
 	columnIndex []*format.ColumnIndex
 	offsetIndex []*format.OffsetIndex
-	encoding    []encoding.Encoding
-	compression []compress.Codec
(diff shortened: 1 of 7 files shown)


column.go
```diff
@@ -303,21 +303,25 @@ func (cl *columnLoader) open(file *File, path []string) (*Column, error) {
...
-				c.encoding = append(c.encoding, LookupEncoding(encoding))
+		if len(c.chunks) > 0 {
+			// Pick the encoding and compression codec of the first chunk.
+			//
+			// Techically each column chunk may use a different compression
+			// codec, and each page of the columm chunk might have a different
+			// encoding. Exposing these details does not provide a lot of value
+			// to the end user.
+			//
+			// Programs that which to determine the encoding and compression of
```
kevinburkesegment, line 314:
```suggestion
			// Programs that wish to determine the encoding and compression of
```

column.go
```diff
@@ -303,21 +303,25 @@ func (cl *columnLoader) open(file *File, path []string) (*Column, error) {
...
 		}
 
-		c.encoding = make([]encoding.Encoding, 0, len(c.chunks))
-		for _, chunk := range c.chunks {
-			for _, encoding := range chunk.MetaData.Encoding {
-				c.encoding = append(c.encoding, LookupEncoding(encoding))
+		if len(c.chunks) > 0 {
+			// Pick the encoding and compression codec of the first chunk.
+			//
+			// Techically each column chunk may use a different compression
```
kevinburkesegment, line 309:
```suggestion
			// Technically each column chunk may use a different compression
```

kevinburkesegment (approved):

PR #145, opened 2022-04-16
improve error handling of invalid arguments in encodings

### LINKED ISSUE #112: ErrInvalidArguments used for both encoding and decoding
Here's ErrInvalidArguments:

```go
	// ErrInvalidArguments is an error returned when arguments passed to the
	// encoding functions are incorrect and will lead to an expected failure.
	//
	// As with ErrNotSupported, this error may be wrapped with specific information
	// about the problem and application are expected to use errors.Is for
	// comparisons
	ErrInvalidArguments = errors.New("invalid encoding arguments")
```

However, this error message gets passed to decoding functions as well as encoding, which can be confusing - I was wondering why we were _encoding_ when we were reading data from a file into a Go struct.

Can we change the error message or use two different error objects?

achille-roussel:
Fixes #112 

I'm taking the suggested approach of making the error messages more descriptive of the conditions that triggered the issues rather than using distinct error values for input validation errors that occur while encoding or decoding.

Let me know if you have any feedback.

### DIFF
--- a/encoding/delta/byte_array_decoder.go
+++ b/encoding/delta/byte_array_decoder.go
@@ -41,7 +41,7 @@ func (d *ByteArrayDecoder) DecodeByteArray(data *encoding.ByteArrayList) (int, e
 
 func (d *ByteArrayDecoder) DecodeFixedLenByteArray(size int, data []byte) (int, error) {
 	if size <= 0 {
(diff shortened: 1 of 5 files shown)


kevinburkesegment (approved):
Thanks!

PR #146, opened 2022-04-18
support pages with unknown bounds

achille-roussel:
This PR modifies the signature of `parquet.Page.Bounds` to return an extra boolean indicating whether the bounds existed. In some cases, page bounds aren't available, for example:
* a page contains no values
* a page contains only null values
* a page was read from a parquet file which contained neither page statistics nor a page index

I've added a new file to the `fixtures` directory, which was part of the original report of an input that parquet-go was failing on https://gist.github.com/senekis/aacc2ea746fcb9da96995ebea009c69e

Please take a look and let me know if you have any concerns.


dictionary.go
```diff
@@ -566,18 +568,20 @@ func (index indexedColumnIndex) NumPages() int       { return 1 }
...
 }
 func (index indexedColumnIndex) MaxValue(int) Value {
-	_, max := index.col.Bounds()
+	_, max, _ := index.col.Bounds()
 	return max
 }
 func (index indexedColumnIndex) IsAscending() bool {
-	return index.col.typ.Compare(index.col.Bounds()) < 0
+	min, max, _ := index.col.Bounds()
+	return index.col.typ.Compare(min, max) <= 0
```
achille-roussel, line 580: @lmarburger I believe this escaped the tests you added in https://github.com/segmentio/parquet-go/pull/138, we might not be validating cases with dictionary pages.

column_reader.go
```diff
@@ -406,8 +417,16 @@ func (r *fixedLenByteArrayColumnReader) ReadValues(values []Value) (n int, err e
 	}
 }
 
-func (r *fixedLenByteArrayColumnReader) Reset(decoder encoding.Decoder) {
+func (r *fixedLenByteArrayColumnReader) Reset(numValues int, decoder encoding.Decoder) {
 	r.decoder = decoder
 	r.buffer = r.buffer[:0]
 	r.offset = 0
+	r.remain = r.size * numValues
```
kevinburkesegment, line 424: A little confusing reading this that `remain` might be the number of items or the number of bytes depending on the struct type.

column_reader_go18.go
```diff
@@ -35,12 +36,16 @@ func (r *columnReader[T]) ReadRequired(values []T) (n int, err error) {
...
 		r.offset += n
+		r.remain -= n
 		values = values[n:]
 	}
-	if r.decoder == nil {
+	if r.remain == 0 || r.decoder == nil {
 		return n, io.EOF
 	}
 	d, err := r.class.decode(r.decoder, values)
+	if r.remain -= d; r.remain == 0 && err == nil {
```
kevinburkesegment, line 46: this is a dumb style thing but this is a lot to read for one line, i'd just do two lines 

```
d, err := ...
r.remain -= d
if r.remain == 0 && err == nil {
    ... 
}
```

dictionary.go
```diff
@@ -73,7 +73,7 @@ func newByteArrayDictionary(typ Type, columnIndex int16, bufferSize int) *byteAr
 		typ: typ,
 		byteArrayPage: byteArrayPage{
 			values:      encoding.MakeByteArrayList(dictCap(bufferSize, 16)),
-			columnIndex: columnIndex,
+			columnIndex: ^columnIndex,
```
kevinburkesegment, line 76: I'm confused why we have an XOR here, at least it's not obvious why a struct property with the same name as the variable needs to be modified by a lot.

encoding/rle/bitpack.go
```diff
@@ -55,31 +63,42 @@ func (d *bitPackRunDecoder) decode(dst []byte, dstWidth uint) (n int, err error)
 		panic("BUG: unsupported destination bit-width")
 	}
 
-	if d.remain -= uint(n); d.remain == 0 {
-		err = io.EOF
-	} else if err != nil {
-		if err == io.EOF {
-			err = io.ErrUnexpectedEOF
+	//fmt.Println("bitPackRunDecoder:", n, err)
```
kevinburkesegment, line 66: Stray? And below too

encoding/rle/decoder.go
```diff
@@ -45,6 +45,15 @@ func (d *Decoder) ReadByte() (byte, error) {
 }
 
 func (d *Decoder) DecodeBoolean(data []bool) (int, error) {
+	//fmt.Println("decode boolean:", len(data), d.decoder)
```
kevinburkesegment, line 48: Stray

file.go
```diff
@@ -469,26 +469,30 @@ type filePages struct {
...
-	if h.IndexPageHeader != nil {
-		h.IndexPageHeader = nil
-	}
-	if h.DictionaryPageHeader != nil {
-		*h.DictionaryPageHeader = format.DictionaryPageHeader{}
-	}
-	if h.DataPageHeaderV2 != nil {
-		*h.DataPageHeaderV2 = format.DataPageHeaderV2{}
-	}
+	/*
```
kevinburkesegment, line 474: Purposeful or not

file.go
```diff
@@ -916,6 +928,14 @@ func (s *filePageValueReaderState) init(columnType Type, column *Column, codec f
 	}
 
 	pageEncoding := pageHeader.Encoding()
+	// In some legacy configuration, the PLAIN_DICTIONARY encoding is used on
```
kevinburkesegment, line 931:
```suggestion
	// In some legacy configurations, the PLAIN_DICTIONARY encoding is used on
```

file_test.go
```diff
@@ -38,28 +42,59 @@ func TestOpenFile(t *testing.T) {
 				t.Errorf("file size mismatch: want=%d got=%d", s.Size(), size)
 			}
 
+			// r := parquet.NewReader(p)
```
kevinburkesegment, line 45: Comments left in

kevinburkesegment (approved):

dictionary.go
achille-roussel: This is a fix for another bug I ran into while debugging the original issue. The reason why we use XOR are documented in https://github.com/segmentio/parquet-go/blob/main/value.go#L39

Technically we could do the XOR later in readers' `ReadValues` method, the alternative here is to compute it ahead of time instead of doing it on each read.

Most of the code was using this approach so I unified on this, we could choose to store the plain value and compute XOR when calling `ReadValues`, it just needs to be consistent across the code base.

encoding/rle/bitpack.go
achille-roussel: I thought I caught them all... thanks!

file.go
achille-roussel: Yes, this code was not used (see comment above), it was confusing to keep it around.

column_reader.go
achille-roussel: True. In this context the `offset` was computed as a number of bytes, so I kept the number of remaining items in bytes as well.

If delaying the scaling from items to bytes will add a significant number of multiplications in the read path 🤷

PR #148, opened 2022-04-22
Simplify row groups APIs

achille-roussel:
I would like to suggest making a change to simplify the `parquet.RowGroup` and `parquet.File` APIs around accessing the lists of row groups and column chunks.

Prior to this change, the abstractions were exposing the lists as pair of methods to get the number of elements, and element at a given index. Programs could iterate over the entries with a loop like `for i, n := 0, f.NumRowGroups(); i < n; i++ { ... }`.

I found that the approach appeared unnecessarily complex from a usability standpoint. Go programs tend to be more expressive if loops can be declared using the `for ... range` construct, but these require exposing the lists as slices.

I am therefore suggesting we change the APIs to expose slices of `parquet.RowGroup` and `parquet.ColumnChunk` values, allowing programs to range over these lists or use the language syntax for slice indexing to access specific elements. It also simplifies the `parquet.RowGroup` implementations since we are removing a method.

The change adds a bit of memory overhead since we are now tracking extra slices of those interface types, I wouldn't expect this to be measurable tho. Technically we would be trading off a bit of computer memory for better usability.

Let me know if you have any concerns.


brancz:
I like this API change a lot!

Pryz (approved):
LGTM ! 🚀

PR #151, opened 2022-04-25
Add more fixture files

achille-roussel:
This PR adds two more files to the list of fixtures that we test against. These were reported to me on Twitter, they highlighted a bug in the `DATE` logical type where we were mistakenly using the `INT64` physical type instead of `INT32`.


lmarburger (approved):

Pryz (approved):

achille-roussel:
Oh nice! Thanks for pointing it out, I'll add these as well.

achille-roussel:
Follow up filed in https://github.com/segmentio/parquet-go/issues/153

Issue #153, opened 2022-04-25
Add parquet-testing files to the fixtures directory


achille-roussel:
https://github.com/apache/parquet-testing/tree/master/data

PR #158, opened 2022-04-26
Add test files

achille-roussel:
This PR adds test files from https://github.com/apache/parquet-testing/tree/master/data to the repository, and addresses the bugs that were highlighted by these tests.

I did not add files for features we don't support yet:
* deprecated Hadoop LZ4 encoding
* encrypted files

There was also one malformed file intended to exercise that the reader was detecting the error, I did not add it since the tests we have are intended to use valid files, testing against malformed input seems like a broader scope change that should be done as a follow up.

Some of the fixes depend on https://github.com/segmentio/encoding/pull/122


kevinburkesegment:
I did not add files for features we don't support yet:

deprecated Hadoop LZ4 encoding
encrypted files

Should we put a note about these in the README?

achille-roussel:
> Should we put a note about these in the README?

Maybe as a follow up, having a "supported features" section in the README could be a good addition (seems like a broader scope than this change tho).

compress/lz4/lz4.go
```diff
@@ -54,55 +42,107 @@ func (c *Codec) CompressionCodec() format.CompressionCodec {
...
+
+	if size := 3 * r.buffer.Len(); cap(r.data) < size {
+		r.data = make([]byte, size)
+	} else {
+		r.data = r.data[:cap(r.data)]
+	}
+
+	for {
+		n, err := lz4.UncompressBlock(r.buffer.Bytes(), r.data)
+		if err != nil {
```
kevinburkesegment, line 105: Check for specific error? This feels fragile

compress/lz4/lz4.go
```diff
@@ -54,55 +42,107 @@ func (c *Codec) CompressionCodec() format.CompressionCodec {
...
+	if r.reader == nil {
+		return io.EOF
+	}
+
+	_, err := r.buffer.ReadFrom(r.reader)
+	if err != nil {
+		return err
+	}
+
+	if size := 3 * r.buffer.Len(); cap(r.data) < size {
```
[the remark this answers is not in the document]
achille-roussel: Optimistic sizing of the output buffer... compression is probably in the 60-80% range, so starting at 3x the input size should minimize the number of loops we need to do.

compress/lz4/lz4.go
achille-roussel: I looked into that, but the `lz4` package doesn't expose the error values (they're in an `internal/lz4errors` package). Based on what I read of the implementation, the only condition where this function errors is if the output buffer is too short.

config.go
```diff
@@ -304,6 +304,16 @@ func SkipPageIndex(skip bool) FileOption {
 	return fileOption(func(config *FileConfig) { config.SkipPageIndex = skip })
 }
 
+// SkipBloomFilters is a file configuration option which when set to true,
+// prevents automatically reading the bloom filters when opening a parquet file.
+// This is useful as an optimization when programs know that they will not need
+// to consume the bloom filters.
+//
+// Defaults to false.
+func SkipBloomFilters(skip bool) FileOption {
```
kevinburkesegment, line 313: Would you ever call this with `false` as the argument?

achille-roussel: I guess if the state is received from user input (e.g. via a command line argument), it's convenient to be able to pass a variable, otherwise the program needs to branch to decide whether to include the option in the option list etc...

kevinburkesegment:
Should we nest all of the test files in a "testdata" directory, that way they won't be included in a vendor directory when you vendor this library

achille-roussel:
Oh I didn't know about `testdata`, I'll rename 👍

achille-roussel:
I filed the documentation follow up in https://github.com/segmentio/parquet-go/issues/159

achille-roussel:
Fixes https://github.com/segmentio/parquet-go/issues/153

PR #161, opened 2022-04-27
Optimize file reader memory footprint

achille-roussel:
This is still a work in progress but generally ready to get feedback I think.

Benchmarks were showing that reading parquet files was allocating large amounts of memory, so I spent a bit of time optimizing it.

I took the following approaches:
* Make the size of internal buffers configurable when reading files, which helps applications reduce the amount of memory they allocate if needed
* Fix a bug that was causing allocations of read buffers to be multiplied by the size of the type (e.g. 8x for a buffer of int64)
* pool the memory buffer used to read compressed pages into memory

These changes have been effective at reducing the amount of memory allocated by the program:
```
name                 old time/op    new time/op    delta
ReaderReadRow/INT64    16.7ns ± 0%    16.1ns ± 1%   -3.81%  (p=0.000 n=9+9)

name                 old speed      new speed      delta
ReaderReadRow/INT64   538MB/s ± 0%   560MB/s ± 1%   +3.96%  (p=0.000 n=9+9)

name                 old row/s      new row/s      delta
ReaderReadRow/INT64     59.8M ± 0%     62.2M ± 1%   +3.96%  (p=0.000 n=9+9)

name                 old alloc/op   new alloc/op   delta
ReaderReadRow/INT64     12.0B ± 0%      1.0B ± 0%  -91.67%  (p=0.000 n=10+10)

name                 old allocs/op  new allocs/op  delta
ReaderReadRow/INT64      0.00           0.00          ~     (all equal)
```

In order to enable pooling of memory buffers, I would like to modify some exported APIs to give the program a mechanism for controlling the lifecycle of some objects (e.g. page readers) via a call to a `Close` method when the objects are not needed anymore. This is an area where I am still experimenting, I may extend this to more APIs to support better internal memory management.

### DIFF
--- a/config.go
+++ b/config.go
@@ -12,6 +12,7 @@ const (
 	DefaultColumnIndexSizeLimit = 16
 	DefaultColumnBufferSize     = 1 * 1024 * 1024
 	DefaultPageBufferSize       = 1 * 1024 * 1024
+	DefaultReadBufferSize       = 4 * 1024
 	DefaultWriteBufferSize      = 32 * 1024
 	DefaultDataPageVersion      = 2
 	DefaultDataPageStatistics   = false
@@ -30,6 +31,7 @@ const (
 //	})
 //
 type FileConfig struct {
+	ReadBufferSize   int
 	SkipPageIndex    bool
 	SkipBloomFilters bool
 }
@@ -38,6 +40,7 @@ type FileConfig struct {
 // default file configuration.
 func DefaultFileConfig() *FileConfig {
 	return &FileConfig{
+		ReadBufferSize:   DefaultReadBufferSize,
 		SkipPageIndex:    DefaultSkipPageIndex,
 		SkipBloomFilters: DefaultSkipBloomFilters,
 	}
@@ -71,7 +74,10 @@ func (c *FileConfig) ConfigureFile(config *FileConfig) {
 
 // Validate returns a non-nil error if the configuration of c is invalid.
 func (c *FileConfig) Validate() error {
-	return nil
+	const baseName = "parquet.(*FileConfig)."
+	return errorInvalidConfiguration(
+		validatePositiveInt(baseName+"ReadBufferSize", c.ReadBufferSize),
+	)
 }
 
 // The ReaderConfig type carries configuration options for parquet readers.
@@ -296,6 +302,14 @@ type WriterOption interface {
 // options for parquet row groups.
 type RowGroupOption interface {
 	ConfigureRowGroup(*RowGroupConfig)
+}
+
+// ReadBufferSize configures the size of the read buffer. This value must be
+// positive.
+//
+// Defaults to 32 KiB.
+func ReadBufferSize(size int) FileOption {
+	return fileOption(func(config *FileConfig) { config.ReadBufferSize = size })
 }
 
 // SkipPageIndex is a file configuration option which prevents automatically
--- a/file.go
+++ b/file.go
@@ -9,6 +9,7 @@ import (
 	"hash/crc32"
 	"io"
 	"sort"
+	"sync"
 
 	"github.com/segmentio/encoding/thrift"
 	"github.com/segmentio/parquet-go/encoding"
@@ -365,7 +366,7 @@ func (g *fileRowGroup) Schema() *Schema                 { return g.schema }
 func (g *fileRowGroup) NumRows() int64                  { return g.rowGroup.NumRows }
 func (g *fileRowGroup) ColumnChunks() []ColumnChunk     { return g.columns }
 func (g *fileRowGroup) SortingColumns() []SortingColumn { return g.sorting }
-func (g *fileRowGroup) Rows() Rows                      { return &rowGroupRowReader{rowGroup: g} }
+func (g *fileRowGroup) Rows() Rows                      { return &rowGroupRows{rowGroup: g} }
 
 type fileSortingColumn struct {
 	column     *Column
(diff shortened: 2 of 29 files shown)


config.go
```diff
@@ -71,7 +74,10 @@ func (c *FileConfig) ConfigureFile(config *FileConfig) {
 
 // Validate returns a non-nil error if the configuration of c is invalid.
 func (c *FileConfig) Validate() error {
-	return nil
+	const baseName = "parquet.(*FileConfig)."
```
kevinburkesegment, line 77: ah my first instinct reading this was that the code was wrong because the function you're calling will always return an error. maybe `checkInvalidConfiguration` ?

file.go
```diff
@@ -455,17 +461,35 @@ type filePages struct {
...
 	// to read whole pages because we have to compute the checksum prior to
 	// exposing the page to the application.
-	compressedPageData []byte
+	compressedPageData *bytes.Buffer
 
 	page filePage
 	skip int64
 }
 
+func (r *filePages) Close() error {
```
kevinburkesegment, line 475: almost wonder about using Reset() as the name here if there's nothing that could return an error

merge.go
```diff
@@ -86,15 +86,22 @@ func (r *mergedRowGroupRowReader) init(m *mergedRowGroup) {
 	}
 }
 
-func (r *mergedRowGroupRowReader) SeekToRow(rowIndex int64) error {
+func (r *mergedRows) Close() error {
+	for i := range r.cursors {
+		r.cursors[i].close()
```
kevinburkesegment, line 91: Check errors?

file.go
kevinburkesegment: I guess just a general comment is I'd expect Close() to actually be associated with some resource that is "open" like a file handle or database connection or something.

It seems we're using it a lot of the time to reclaim buffer space.

As someone who's not super familiar, it makes it pretty difficult to reason about, what is actually happening when I call Close() on a given resource.

page.go
```diff
@@ -777,46 +648,30 @@ func (page *fixedLenByteArrayPage) WriteTo(e encoding.Encoder) error {
...
-	page   *fixedLenByteArrayPage
-	offset int
-}
-
-func (r *fixedLenByteArrayPageReader) Read(b []byte) (n int, err error) {
-	n, err = r.ReadFixedLenByteArrays(b)
-	return n * r.page.size, err
-}
+const (
+	bufferSize = 4096
```
kevinburkesegment, line 658: maybe just me but I always appreciate a note about why we chose a particular value; can help educate about what you think the constraint is that's worth optimizing for.

page_decoder.go
```diff
@@ -0,0 +1,399 @@
...
+			}
+			if len(repetitionLevels) < numValues {
+				numValues = len(repetitionLevels)
+			}
+		}
+
+		if r.maxDefinitionLevel > 0 {
+			definitionLevels, err = r.definitions.peekLevels()
+			if err != nil {
+				return read, fmt.Errorf("decoding definition level from data page of column %d: %w", ^r.columnIndex, err)
```
kevinburkesegment, line 77: this bit flipping is a trip every time I read it. wondering if we should just make this a function that would get inlined? 

make the variable name `_columnIndex` and then `r.columnIndex()` returns the value or something?

kevinburkesegment: Like if I had to work with this codebase I guarantee I would make a mistake while using columnIndex by not doing the right thing.

page_decoder.go
```diff
@@ -0,0 +1,399 @@
...
+			return n, err
+		}
+
+		r.buffer = buffer[:d*r.size]
+		r.offset = 0
+	}
+}
+
+var (
+	byteArrayLists sync.Pool // *encoding.ByteArrayList
```
kevinburkesegment, line 369: These comments are super helpful, thanks.

page_go18.go
```diff
@@ -87,22 +95,103 @@ func (r *pageReader[T]) ReadRequired(values []T) (n int, err error) {
...
+		remain:      numValues,
+		columnIndex: ^columnIndex,
+	}
+}
+
+func (r *pageDecoder[T]) Type() Type {
+	return r.typ
+}
+
+func (r *pageDecoder[T]) Column() int {
```
kevinburkesegment, line 146: Ah, yeah this is good. Can we do this in more places?

kevinburkesegment (approved):

page_decoder.go
achille-roussel: I agree this is tricky to get right, I'll revisit 👍

file.go
achille-roussel: I've added `Close` methods on types that act as streams/readers, for example in this case the `parquet.Pages` interface is defined as:
```go
type Pages interface {
    PageReader
    io.Closer
}

type PageReader interface {
    ReadPage() (Page, error)
}
```
I'm under the impression that this is a common pattern, it reminds of `io.ReadCloser` for example. Closing the reader helps release resources, but it also acts as a mechanism to interrupt the stream, where after a call to `Close` the reads will return `io.EOF`.

What do you think?

page_decoder.go
achille-roussel: Thanks! I took this habit because `sync.Pool` really says nothing about the objects that are expected to be stored in it, and the APIs are untyped so the compiler doesn't help either.

When we drop support for Go 1.17 we could put a generic API on `sync.Pool` and have `syncPool[*encoding.ByteArrayList]` which will be even better.

achille-roussel:
Closing in favor of https://github.com/segmentio/parquet-go/pull/182

PR #163, opened 2022-04-29
Refactor compression APIs

achille-roussel:
As I was working on https://github.com/segmentio/parquet-go/pull/161, it became apparent that a lot of the complexity of reading files came from the way the compression APIs were based on reading and writing byte streams, despite pages always being manipulated in memory.

With this PR, I am attempting to reduce this complexity to enable further optimizations by changing the compression models to stateless `Encode`/`Decode` methods instead of the `NewReader`/`NewWriter` we had previously.

While this is a breaking change, I don't expect this to impact users much since the compression codecs are managed internally, programs do not need to use them directly.

### DIFF
--- a/compress/lz4/lz4.go
+++ b/compress/lz4/lz4.go
@@ -52,6 +52,9 @@ func (c *Codec) Encode(dst, src []byte) ([]byte, error) {
 }
 
 func (c *Codec) Decode(dst, src []byte) ([]byte, error) {
+	// 3x seems like a common compression ratio, so we optimistically size the
+	// output buffer to that size. Feel free to change the value if you observe
+	// different behaviors.
 	dst = reserveAtLeast(dst, 3*len(src))
 
 	for {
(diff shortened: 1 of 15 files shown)


compress/lz4/lz4.go
```diff
@@ -41,115 +37,44 @@ func (c *Codec) CompressionCodec() format.CompressionCodec {
...
-	}
-
-	optimisticOutputSize := 3 * r.buffer.Len()
-	if cap(r.data) < optimisticOutputSize {
-		r.data = make([]byte, optimisticOutputSize)
-	} else {
-		r.data = r.data[:cap(r.data)]
-	}
+func (c *Codec) Decode(dst, src []byte) ([]byte, error) {
+	dst = reserveAtLeast(dst, 3*len(src))
```
kevinburkesegment, line 55: I guess I'd put a comment in here, like, "observation shows that decoded bytes are 3x the size of encoded bytes" or something, or just, an answer to the question "how did you pick the number three"

kevinburkesegment (approved):

PR #169, opened 2022-05-03
example_test: add package example

kevinburkesegment:
This is a very simple example of how to roundtrip data from a struct
to a file and back.

### DIFF
--- a/example_test.go
+++ b/example_test.go
@@ -1,0 +1,61 @@
+package parquet_test
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"log"
+	"os"
+
+	"github.com/segmentio/parquet-go"
+)
+
+func Example() {
+	// parquet-go uses the same struct-tag definition style as JSON and XML
+	type Contact struct {
+		Name string `parquet:"name"`
+		// "zstd" specifies the compression for this column
+		PhoneNumber string `parquet:"phoneNumber,optional,zstd"`
+	}
+
+	type AddressBook struct {
+		Owner             string    `parquet:"owner,zstd"`
+		OwnerPhoneNumbers []string  `parquet:"ownerPhoneNumbers,gzip"`
+		Contacts          []Contact `parquet:"contacts"`
+	}
+
+	f, _ := ioutil.TempFile("", "parquet-example-")
+	writer := parquet.NewWriter(f)
+	rows := []AddressBook{
+		{Owner: "UserA", Contacts: []Contact{
+			{Name: "Alice", PhoneNumber: "+15505551234"},
+			{Name: "Bob"},
+		}},
+		// Add more rows here.
+	}
+	for _, row := range rows {
+		if err := writer.Write(row); err != nil {
+			log.Fatal(err)
+		}
+	}
+	_ = writer.Close()
+	_ = f.Close()
+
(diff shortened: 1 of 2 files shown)


file.go
```diff
@@ -136,19 +136,19 @@ func OpenFile(r io.ReaderAt, size int64, options ...FileOption) (*File, error) {
 // Only leaf columns have indexes, the returned indexes are arranged using the
 // following layout:
 //
-//	+ -------------- +
+//	---------------- +
```
kevinburkesegment, line 139: Go 1.19 adds some documentation formatting/rewrites. It wasn't happy with the initial layout here and kept trying to reformat it. It's happy with this format, if you don't mind it.

achille-roussel: Could we change the trailing ` +` into `--` as well for consistency?

example_test.go
```diff
@@ -0,0 +1,60 @@
...
+			{Name: "Bob"},
+		}},
+		// Add more rows here.
+	}
+	for _, row := range rows {
+		if err := writer.Write(row); err != nil {
+			log.Fatal(err)
+		}
+	}
+	_ = writer.Close()
```
achille-roussel, line 40: Error handling seems to be a bit inconsistent here, we check on `Write` but not on `Close`, which might error since it writes the file footer etc... Maybe the example encourages the wrong practice here, how about checking the error returned by `Close`?

example_test.go
```diff
@@ -0,0 +1,60 @@
...
+	_ = f.Close()
```
achille-roussel, line 41: Is the `_ =` intended to silence linter rules?

example_test.go
```diff
@@ -0,0 +1,60 @@
...
+
+	// Now, we can read from the file.
+	rf, _ := os.Open("testdata/data.parquet")
```
achille-roussel, line 44: How about using `/tmp` instead of testdata? Otherwise running this example will add a new file to the repository that will likely end up being checked in by accident.

kevinburke: Ah, it does but we have the testdata directory checked in to gitignore currently. OK

achille-roussel: 🤔 

That seems unexpected, we have plenty of files there in the repo https://github.com/segmentio/parquet-go/tree/main/testdata, that gitignore rule may not be stated properly and should be removed.

example_test.go
kevinburkesegment: Yeah sorry was trying to strike a balance between brevity and explanation

example_test.go
kevinburkesegment: I wanted to indicate this function can return an error that you should check without adding 2 lines to the example to check it

example_test.go
kevinburkesegment: https://github.com/segmentio/parquet-go/pull/171

file.go
kevinburkesegment: Done

example_test.go
achille-roussel: My opinion is we should do the error check, younger engineers will use these examples as "good practice" without really thinking through the consequences of not checking the errors, we will serve them better by not teaching them to ignore errors I think.

mihatfield (approved):
lgtm

Issue #154, opened 2022-04-25
Support deprecated BIT_PACKED encoding


achille-roussel:
The parquet file format used to support a BIT_PACKED encoding which represented bits in MSB order, but has since been deprecated.

To support reading files that were encoded with columns in this format, we should add support in parquet-go.

Specification:
* https://github.com/apache/parquet-format/blob/master/Encodings.md#bit-packed-deprecated-bit_packed--4

PR #172, opened 2022-05-05
Refactor encoding APIs (part 1)

achille-roussel:
This PR is a follow up to the work started in #163, focusing on the `encoding` package this time.

The core change consists in changing the design of encodings from using streams to working on in-memory buffers. This change simplifies the implementations, which enables better memory management. In most cases the users of encodings can "bring their own buffers", allowing them full control over the memory footprint and management.

I am stating this PR as a "part 1" because there are more changes I want to make, but it is already a significant amount of work so now seems like a good time to land a first milestone.

I did not invest a lot of time in measuring performance of the new code, this seemed premature as the goal here was to focus on the refactor. There may be performance regressions introduced by this change, but some code paths may also have become more efficient. Performance optimization will be done in a follow up PR as well.

The test suites we have were really effective at detecting regressions during the refactor, which gives me confidence that we may be able to land this without introducing damaging bugs, I wouldn't expect code quality to decrease.

This change will also enables the work on #154, as implementing encodings is now a lot simpler due to now having to implement stream APIs.


bloom.go
```diff
@@ -124,128 +131,128 @@ func searchBloomFilterColumn(filters []BloomFilterColumn, path columnPath) Bloom
...
-
-func newBloomFilterEncoder(filter bloom.MutableFilter, hash bloom.Hash) *bloomFilterEncoder {
-	return &bloomFilterEncoder{filter: filter, hash: hash}
-}
+const (
+	// Size of the stack buffer used to perform bulk operations on bloom filters.
+	//
+	// This value was determined as being a good default empirically,
+	// 128 x uint64 makes a 1KiB buffer which amortizes the cost of calling
+	// methods of bloom filters while not causing too much stack growth either.
```
kevinburkesegment, line 139: 👍

bloom.go
```diff
@@ -124,128 +131,128 @@ func searchBloomFilterColumn(filters []BloomFilterColumn, path columnPath) Bloom
...
-	return e.insert8(bits.BoolToBytes(data))
+func (splitBlockEncoding) EncodeInt64(dst []byte, src []int64) ([]byte, error) {
+	splitBlockEncodeUint64(bloom.MakeSplitBlockFilter(dst), bits.Int64ToUint64(src))
+	return dst, nil
 }
 
-func (e *bloomFilterEncoder) EncodeInt8(data []int8) error {
-	return e.insert8(bits.Int8ToBytes(data))
+func (e splitBlockEncoding) EncodeInt96(dst []byte, src []deprecated.Int96) ([]byte, error) {
+	splitBlockEncodeFixedLenByteArray(bloom.MakeSplitBlockFilter(dst), deprecated.Int96ToBytes(src), 12)
```
kevinburkesegment, line 163: Missed conversion of deprecated to bits here?

column.go
```diff
@@ -471,6 +473,283 @@ func schemaRepetitionTypeOf(s *format.SchemaElement) format.FieldRepetitionType
...
+
+func (p *dataPage) decompress(codec compress.Codec, data []byte) (err error) {
+	p.values, err = codec.Decode(p.values, data)
+	p.data, p.values = p.values, p.data[:0]
+	return err
+}
+
+func (p *dataPage) decode(typ Type, enc encoding.Encoding, data []byte) error {
+	// Note: I am not sold on this design, it parts ways from the way type
+	// specific behavior are implemented in other places based on the Type
```
kevinburkesegment, line 496:
```suggestion
	// specific behavior is implemented in other places based on the Type
```

encoding/fuzz_test.go
```diff
@@ -410,3 +411,4 @@ func (s *byteSource) Int63() int64 {
 }
 
 func (s *byteSource) Seed(seed int64) {}
+*/
```
kevinburkesegment, line 414: Deliberate?

encoding/notsupported.go
```diff
@@ -27,151 +26,168 @@ var (
 	ErrInvalidArgument = errors.New("invalid argument")
 )
 
-// NotSupported is a type satisfying the Encoding interface which does not
-// support encoding nor decoding any value types.
-type NotSupported struct {
+// Error constructs an error which wraps err and indicate that it originated
```
kevinburkesegment, line 29:
```suggestion
// Error constructs an error which wraps err and indicates that it originated
```

encoding/notsupported.go
```diff
@@ -27,151 +26,168 @@ var (
...
-	return false
+// ErrInvalidInputSize constructs an error indicating that decoding failed due
+// to the size of the input.
+func ErrInvalidInputSize(e Encoding, typ string, size int) error {
+	return Errorf(e, "cannot decode %s from input of size %d: %w", typ, size, ErrInvalidArgument)
 }
 
-func (NotSupported) NewDecoder(io.Reader) Decoder {
-	return NotSupportedDecoder{}
+// CanEncodeBoolean returns true if the e can encode BOOLEAN values.
```
kevinburkesegment, line 47:
```suggestion
// CanEncodeBoolean reports whether e can encode BOOLEAN values.
```

writer.go
```diff
@@ -664,11 +655,102 @@ func (w *writer) WritePage(page Page) (int64, error) {
 	return w.columns[page.Column()].WritePage(page)
 }
 
+// One writerBuffers is used by each writer instance, the memory buffers here
+// are shared by all columns of the writer because serialization is not done
+// concurrently, which helps keep memory utilization low, both it the total
```
kevinburkesegment, line 660:
```suggestion
// concurrently, which helps keep memory utilization low, both in the total
```

writer.go
```diff
@@ -664,11 +655,102 @@ func (w *writer) WritePage(page Page) (int64, error) {
...
+// footprint and GC cost.
+//
+// The type alose exposes helper methods to facilitate the generation of parquet
```
kevinburkesegment, line 663: Not sure what "alose" means here

writer.go
```diff
@@ -732,14 +809,17 @@ func (c *writerColumn) reset() {
...
 	for i := range c.pages {
 		c.pages[i] = nil
 	}
-	c.filter = c.filter[:0]
+	for i := range c.filter.pages {
+		c.filter.pages[i] = nil
+	}
 	c.pages = c.pages[:0]
+	// Bloom filters may change in size between row groups, but we retain the
+	// buffer to avoir reallocating large memory blocks.
```
kevinburkesegment, line 820:
```suggestion
	// buffer to avoid reallocating large memory blocks.
```

writer.go
```diff
@@ -1198,23 +1239,34 @@ func (c *writerColumn) writeDictionaryPage(output io.Writer, dict Dictionary) er
...
+			c.pool.PutPageBuffer(buffer)
 		}
-		pageData = b
+	}()
+	written, err := writeTo(buffer)
+	if err != nil {
+		return err
+	}
+	if written != size {
+		return fmt.Errorf("writing parquet colum page expected %dB but got %dB: %w", size, written, io.ErrShortWrite)
```
kevinburkesegment, line 1266:
```suggestion
		return fmt.Errorf("writing parquet column page expected %dB but got %dB: %w", size, written, io.ErrShortWrite)
```

bloom.go
kevinburkesegment: The other functions above and below reference `bits` instead of `deprecated` which made me think you just missed this line while converting the API's above and below.

achille-roussel: Ah, that's because the `deprecated` package contains everything related to the `deprecated.Int96` type, `bits` has things for all other types.

encoding/fuzz_test.go
achille-roussel: The tests need to be adapted to the new interfaces, they aren't valid anymore but I think they are useful to keep around as baseline to be reworked in a follow up PR.

achille-roussel: https://github.com/segmentio/parquet-go/pull/174/files

achille-roussel:
I'm on the fence of adding a CHANGELOG:
* we're still early so a lot is changing and the value of tracking it is low
* github is going to generate change logs automatically when we will start versioning the project

I will think more about it.

achille-roussel:
Thanks for the review!

PR #178, opened 2022-05-16
refactor indexed page type

achille-roussel:
Follow up to #177, this PR refactors the `indexedType` by splitting it in `indexedType` and `indexedPageType`, the latter providing the override of `Encode`/`Decode` intended to support serializing and deserializing indexed pages.

The split is more meaningful in my opinion because the previous form was somewhat inconsistent where the Type returned by a dictionary would `Encode`/`Decode` methods that did not match the actual type of values held in the dictionary.


dictionary.go
```diff
@@ -919,41 +917,36 @@ func newIndexedType(typ Type, dict Dictionary) *indexedType {
...
-	return enc.EncodeInt32(dst, src)
-}
-
-func (t *indexedType) Decode(dst, src []byte, enc encoding.Encoding) ([]byte, error) {
-	return enc.DecodeInt32(dst, src)
+	return newIndexedPage(t, makeColumnIndex(columnIndex), makeNumValues(numValues), data)
 }
 
+// indexedPage is an implementation of the BufferedPage interface which stores
+// indexes instead of plain value. The indexes reference the value values in a
```
kevinburkesegment, line 928:
```suggestion
// indexes instead of plain value. The indexes reference the values in a
```

kevinburkesegment (approved):

PR #179, opened 2022-05-16
fix bit-packed encoding of booleans in memory

achille-roussel:
Following up from #178, this PR modifies the encoding of boolean values in memory to be bit-packed instead of holding one value per byte.

The intent is to unify the implementation of parquet pages to always use the PLAIN encoding in memory:
* when calling `parquet.Page.Data()` on boolean pages
* when constructing boolean pages with `parquet.Type.NewPage`
* when constructing boolean dictionaries with `parquet.Type.NewDictionary`

As a side effect, the memory footprint of boolean pages is reduced by 87.5% since we are holding 8 values per byte instead of 1.


column_buffer.go
```diff
@@ -684,7 +684,7 @@ func newBooleanColumnBuffer(typ Type, columnIndex int16, bufferSize int) *boolea
 	return &booleanColumnBuffer{
 		booleanPage: booleanPage{
 			typ:         typ,
-			values:      make([]bool, 0, bufferSize),
+			values:      make([]byte, 0, bufferSize),
```
kevinburkesegment, line 687: The contract is implicitly changing here, since callers will be able to store 8x as many values in `bufferSize` as they could previously, I guess it's ok though.

kevinburkesegment: I'm also nervous about someone attempting to read/write `values` directly, in Python or whatever I'd put `_values` to make that clear, but probably unnecessary here.

column_buffer.go
```diff
@@ -716,46 +718,73 @@ func (col *booleanColumnBuffer) Pages() Pages { return onePage(col.Page()) }
...
-	col.values[i], col.values[j] = col.values[j], col.values[i]
+func (col *booleanColumnBuffer) valueAt(i int) bool {
+	j := uint32(i) / 8
+	k := uint32(i) % 8
+	return ((col.values[j] >> k) & 1) != 0
 }
 
-func (col *booleanColumnBuffer) Write(b []byte) (int, error) {
-	return col.WriteBooleans(bits.BytesToBool(b))
+func (col *booleanColumnBuffer) setValueAt(i int, v bool) {
```
kevinburkesegment, line 743: Thread safety issue?

dictionary.go
```diff
@@ -934,10 +939,28 @@ type indexedPage struct {
 }
 
 func newIndexedPage(typ *indexedType, columnIndex int16, numValues int32, values []byte) *indexedPage {
-	values = resize(values, 4*int(numValues))
+	// RLE encoded values that contains dictionary indexes in data pages are
```
kevinburkesegment, line 942:
```suggestion
	// RLE encoded values that contain dictionary indexes in data pages are
```

encoding/rle/rle.go
```diff
@@ -95,6 +95,54 @@ func (e *Encoding) wrap(err error) error {
...
+			if n := j - i; n > 1 {
+				dst = appendUvarint(dst, uint64(8*n)<<1)
+				dst = append(dst, src[i])
+				i = j
+				continue
+			}
+		}
+
+		// Sequences of bits that are neither all zeroes or ones are bit-packed,
+		// which is a simple copy of the input to the output preceeded with the
```
kevinburkesegment, line 127:
```suggestion
		// which is a simple copy of the input to the output preceded with the
```

kevinburkesegment (approved):

column_buffer.go
achille-roussel: The ColumnBuffer types are not intended to be used concurrently from multiple goroutines so I believe we're ok here, unless there was something else you wanted to point out?

column_buffer.go
achille-roussel: The semantics don't change since the buffer size is intended to be expressed in bytes. It's true that the buffer can hold 8x more values for the same space then.

achille-roussel: I'm going to rename from `values` to `bits` to be more explicit that we are bit-packing the booleans 👍

achille-roussel:
Thanks for the reviews!

Issue #135, opened 2022-04-13
Caching parquet metadata in an off-process cache


annanay25:
As discussed earlier, in Grafana Tempo we are planning to build an object store backed parquet database. At higher volumes, querying could mean retrieving metadata from a large number of parquet files and we would like to optimise by caching these in an off-process cache (redis/memcached/...)

There are two strategies we have considered to implement this:
1. Build in support for metadata caching directly into this library so the details are abstracted away from the user. 
2. Provide methods to marshal/unmarshal metadata and handle all the caching on the application side.

Thoughts?

Using this ticket as a placeholder to start discussions around this.

mdisibio:
Adding a few additional thoughts: 

1. The library currently takes an `io.ReaderAt` which we already wrap for various reasons (metrics, etc).  We considered if there is a way to redirect some reads to cache but did not see a general solution there.
2. We expect our largest blocks to be 20+GB and it looks like metadata for them is 50-100MB.  This is quite large for a single cache entry, so it might be best to support granularity of individual row groups, or even column chunks which allows for reading the minimum necessary data for the columns involved.

achille-roussel:
Do you have more details to share about what metadata you would like to optimize access to?

annanay25:
I believe we would need to cache the entire metadata in order to avoid hitting the object storage to fetch anything other than data? Going through code I think we use Schema and RowGroup information from the reader, but even so we might want to cache the entire metadata

achille-roussel:
OK, I was wondering if you intended to also cache other parts of the file like column indexes or bloom filters.

> The library currently takes an io.ReaderAt which we already wrap for various reasons (metrics, etc). We considered if there is a way to redirect some reads to cache but did not see a general solution there.

This would be hard to do without duplicating the logic to decode footers of parquet files. A lower-level caching solution for the `io.ReaderAt` layer would be more appropriate in my opinion.

> We expect our largest blocks to be 20+GB and it looks like metadata for them is 50-100MB. This is quite large for a single cache entry, so it might be best to support granularity of individual row groups, or even column chunks which allows for reading the minimum necessary data for the columns involved.

This is really large indeed, almost questionable whether caching of metadata would improve access here: assuming the data is on some long term storage medium, loading large blocks will be bound on I/O bandwidth, not latency, so it may be just as efficient to read the original files.

> Build in support for metadata caching directly into this library so the details are abstracted away from the user.

[...]

PR #180, opened 2022-05-16
Add support for caching parquet metadata

annanay25:
As per our discussion in https://github.com/segmentio/parquet-go/issues/135, opening this draft PR to support caching parquet metadata.

The PR includes some plumbing to ensure the right options are supplied to the parquet reader/writers, allowing them to use secondary `io.ReaderAts` to fetch column chunk data. 

Marking draft to ensure we are aligned on the general design and direction of implementation.


config.go
```diff
@@ -30,8 +31,9 @@ const (
 //	})
 //
 type FileConfig struct {
-	SkipPageIndex    bool
-	SkipBloomFilters bool
+	SkipPageIndex       bool
+	SkipBloomFilters    bool
+	GetIOReaderFromPath func(filepath string) io.ReaderAt
```
achille-roussel, line 36: What do you think about using an interface type here instead of a function?

config.go
```diff
@@ -134,6 +136,7 @@ func (c *ReaderConfig) Validate() error {
 //
 type WriterConfig struct {
 	CreatedBy            string
+	ColumnChunkFilePath  string
```
achille-roussel, line 139: Maybe this could be an interface which defines both the file path and the location where the writer would output the metadata? (e.g. a Name method and io.Writer so we could use a os.File in the simplest cases)

annanay25: Ok I removed the option of writing metadata to a different location. I realise we don't need this if we are only looking at caching metadata/indexes. We can make use of secondary (cache-free) io readers while reading column chunk data

README.md
```diff
@@ -430,6 +430,59 @@ Reading arrays of typed values is often preferable when performing aggregations
 on the values as this model offers a more compact representation of the values
 in memory, and pairs well with the use of optimizations like SIMD vectorization.
 
+#### A. Caching parquet file metadata
```
annanay25, line 433: Still need to update this and write a test!

config.go
annanay25: Done, but I'm not sure about what the initial signature should look like. The application should have enough information about the object to provide a secondary reader, so I'm not even sure we need to pass the column chunk metadata

annanay25:
@achille-roussel pinging on this one!

achille-roussel:
> Ok I removed the option of writing metadata to a different location. I realise we don't need this if we are only looking at caching metadata/indexes. We can make use of secondary (cache-free) io readers while reading column chunk data

I'm curious how the metadata end up being cached if the writer doesn't produce them, are you using a read-through cache then?

annanay25:
Yes that's right, we are using a read-through. Are there any concerns with regard to that?

annanay25:
I'm also thinking if it is good design to add an override for reading columnChunk data through a secondary reader... Would it be better design to add an override for reading metadata through the cached reader? That way we can incrementally add caching support for metadata objects to improve speed vis-a-vis the current situation where we could potentially end up caching large objects unintentionally

achille-roussel:
Regarding reading column data, I wrote down some thoughts in https://github.com/segmentio/parquet-go/issues/239; when you get a chance to read through, let me know what you think about it!

annanay25:
👍 will take a look, thanks! Do you think we need something more to proceed
with this PR?

PR #182, opened 2022-05-17
Optimize file reader

achille-roussel:
This PR ports the changes from #161 onto #175.

Thanks to the couple rounds of refactors, this change is now a lot simpler to land and the optimizations related to more efficient memory management are also much simpler.

I tried to capture context in docs and comments in the code, please let me know if you have any feedback!

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -24,6 +24,9 @@ import (
 //		}
 //		rows = append(rows, row)
 //	}
+//	if err := reader.Close(); err != nil {
+//		...
+//	}
 //
 //
 type Reader struct {
--- a/reader.go
+++ b/reader.go
@@ -283,7 +283,7 @@ func (r *reader) Reset() {
 		// This optimization works for the common case where the underlying type
 		// of the Rows instance is rowGroupRows, which should be true in most
 		// cases since even external implementations of the RowGroup interface
-		// can constructs values of this type via the NewRowGroupRowReader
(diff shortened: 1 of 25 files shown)


kevinburkesegment (commented):
Why can't we call Close() when we hit EOF on the Reader?

reader.go
```diff
@@ -259,11 +277,28 @@ func (r *reader) init(schema *Schema, rowGroup RowGroup) {
...
 
 func (r *reader) Reset() {
-	r.rows = nil // TODO: can we make the RowReader reusable?
 	r.rowIndex = 0
+
+	if rows, _ := r.rows.(*rowGroupRows); rows != nil {
+		// This optimization works for the common case where the underlying type
+		// of the Rows instance is rowGroupRows, which should be true in most
+		// cases since even external implementations of the RowGroup interface
+		// can constructs values of this type via the NewRowGroupRowReader
```
kevinburkesegment, line 286:
```suggestion
		// can construct values of this type via the NewRowGroupRowReader
```

achille-roussel:
> Why can't we call Close() when we hit EOF on the Reader?

The `Reader` may be reused via a call to `Reset` or `SeekToRow`, if we call `Close` we require re-creating the underlying `Rows` objects. This wouldn't be wrong, but we would miss the opportunity to more efficiently reuse the resources we already have.

kevinburkesegment (approved):

Issue #149, opened 2022-04-23
Design improvements for row readers and writers


achille-roussel:
In this issue, I would like to submit a proposal for modifying the `parquet.RowReader` and `parquet.RowWriter` interfaces in order to improve usability of those abstractions, as well as reducing their overhead to improve application performance.

# Context

At this time, the APIs in question have the following definitions:

```go
package parquet 

type Row []Value

type RowReader interface {
    ReadRow(Row) (Row, error)
}

type RowWriter interface {
    WriteRow(Row) error
}
```

## Qualities

The goals of these APIs is to allow the expression of generic algorithms working on sequences of parquet rows. For example, when writing rows to a `parquet.Writer`, applications can call `WriteRow` repeatedly to append rows and construct a parquet file. When reading rows, the API allows for a stream-like consumption of the content of a parquet file, which is useful when the datasets cannot be loaded entirely in memory (a common case when working with large parquet files).

### Simplicity of the design

The use of `parquet.Row` offers a simple abstraction (a slice doesn't hide much about the implementation details), and enables constructs that leverage language features like `append` or `for ... range`. For example, some programs can simply iterate over the values of a row with a construct like this:
```go
for _, value := range row {
    ...
}
```

### Similarities with the `io` package

Being somewhat similar to the design of `io.Reader` and `io.Writer`, this model also reuses concepts that Go developers are often familiar with, helping ramp up when learning to use the package. The parallels go beyond the interfaces into APIs like `parquet.CopyRows` which mimic the design of `io.Copy`, etc... creating a strong sense of familiarity for developers who have worked with the standard `io` package.

## Flaws

There are limitations with the design of these APIs that seem to have a greater impact than originally anticipated, and we might be able to mitigate by striking different balances in the design.

### Asymmetry of parquet.RowReader and parquet.RowWriter

When writing rows to a `parquet.RowWriter`, the application forms a full `parquet.Row` and passes it to the writer, allowing the `WriteRow` method to offer very simple input and output. The _intuitive_ corresponding API for `parquet.RowReader` would be to consume and return the next row, for example:
```go
interface {
    ReadRow() (Row, error)
}
```

Performance consideration arise with this model, since the `parquet.Row` type needs to be somewhat flexible in the amount of memory that it consumes (e.g. repeated columns may cause the number of values in a row to vary). Returning a newly allocated `parquet.Row` value on every call would greatly increase GC pressure on the application, as well as impact the compute footprint due to extra malloc calls.

[...]

PR #160, opened 2022-04-26
Refactor row reader/writer APIs

achille-roussel:
This PR is a MVP version of the proposal in https://github.com/segmentio/parquet-go/issues/149, demonstrating how the changes would modify the APIs.

PR #183, opened 2022-05-17
Refactor parquet.RowReader and parquet.RowWriter

achille-roussel:
Based on #175, this PR ports the changes from #160 on top of the recent refactors.

I added benchmarks for the `parquet.Buffer` type that would highlight the effect of the API design change. Here are some preliminary results from the design change:
```
$ benchstat /tmp/BenchmarkBufferReadRows100x.{0,1}
name                old time/op    new time/op    delta
BufferReadRows100x    1.67µs ± 1%    1.59µs ± 0%  -4.85%  (p=0.000 n=10+9)

name                old row/s      new row/s      delta
BufferReadRows100x     59.8M ± 0%     62.8M ± 0%  +5.03%  (p=0.000 n=9+9)

name                old alloc/op   new alloc/op   delta
BufferReadRows100x     0.00B          0.00B         ~     (all equal)

name                old allocs/op  new allocs/op  delta
BufferReadRows100x      0.00           0.00         ~     (all equal)
```
```
$ benchstat /tmp/BenchmarkBufferWriteRows100x.{0,1}
name                 old time/op    new time/op    delta
BufferWriteRows100x    3.79µs ± 0%    1.41µs ± 0%   -62.82%  (p=0.000 n=9+9)

name                 old row/s      new row/s      delta
BufferWriteRows100x     26.4M ± 0%     70.9M ± 0%  +168.93%  (p=0.000 n=9+9)

name                 old alloc/op   new alloc/op   delta
BufferWriteRows100x     0.00B          0.00B           ~     (all equal)

name                 old allocs/op  new allocs/op  delta
BufferWriteRows100x      0.00           0.00           ~     (all equal)
```

The gains are not as significant on the read path because the inner implementations are not yet optimized to leverage the new APIs, but it's nice to see how much efficiency we enabled on the write path with it.

Fixes https://github.com/segmentio/parquet-go/issues/149

### DIFF
--- a/reader.go
+++ b/reader.go
@@ -24,6 +24,9 @@ import (
 //		}
 //		rows = append(rows, row)
 //	}
+//	if err := reader.Close(); err != nil {
+//		...
+//	}
 //
 //
 type Reader struct {
--- a/reader.go
+++ b/reader.go
@@ -189,7 +189,8 @@ func (r *Reader) Read(row interface{}) error {
 		return fmt.Errorf("seeking reader to row %d: %w", r.rowIndex, err)
 	}
 
-	if _, err := r.read.ReadRows(r.values[:]); err != nil {
+	n, err := r.read.ReadRows(r.values[:])
+	if n == 0 {
 		return err
 	}
 
@@ -276,13 +277,16 @@ func (r *reader) init(schema *Schema, rowGroup RowGroup) {
 func (r *reader) Reset() {
 	r.rowIndex = 0
 
-	if rows, _ := r.rows.(*rowGroupRows); rows != nil {
+	if rows, ok := r.rows.(interface{ Reset() }); ok {
 		// This optimization works for the common case where the underlying type
 		// of the Rows instance is rowGroupRows, which should be true in most
 		// cases since even external implementations of the RowGroup interface
 		// can construct values of this type via the NewRowGroupRowReader
 		// function.
-		rows.reset()
+		//
+		// Foreign implementations of the Rows interface may also define a Reset
+		// method in order to participate in this optimization.
+		rows.Reset()
 		return
 	}
 
--- a/reader.go
+++ b/reader.go
@@ -1,6 +1,7 @@
 package parquet
 
 import (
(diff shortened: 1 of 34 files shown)


reader.go
```diff
@@ -34,7 +34,7 @@ type Reader struct {
 	file     reader
 	read     reader
 	rowIndex int64
-	values   []Value
+	values   [1]Row
```
achille-roussel, line 37: Following up from https://github.com/segmentio/parquet-go/pull/160/files#r861234029

I'm going to keep it as-is for now because having an array helps avoid creating temporary slices when calling `ReadRows`, will revisit later when we work on optimizing the read path.

reader_test.go
```diff
@@ -369,14 +369,14 @@ func BenchmarkReaderReadRow(b *testing.B) {
 			}
 
 			r := parquet.NewReader(f)
-			rowbuf := make(parquet.Row, 0, 16)
+			rowbuf := make([]parquet.Row, 0, 16)
```
achille-roussel, line 372: Following up from https://github.com/segmentio/parquet-go/pull/160/files#r861234701

`defaultRowBufferSize` is in the `parquet` package, this is a test so I think it's OK to keep as-is.

row.go
```diff
@@ -55,7 +60,18 @@ type RowSeeker interface {
...
-	ReadRow(Row) (Row, error)
+	// Read rows from the reader, returning the number of rows read into the
+	// buffer, and any error that occurred.
+	//
+	// When all rows have been read, the reader returns io.EOF to indicate the
+	// end of the sequence. It is valid for the reader to return both a non-zero
+	// number of rows and a non-nil error (including io.EOF).
+	//
+	// The buffer of rows passed as argument will be used to store values of
+	// each row read from the reader. If the rows are no nil, the backing array
```
kevinburkesegment, line 71:
```suggestion
	// each row read from the reader. If the rows are not nil, the backing array
```

kevinburkesegment (approved):

PR #186, opened 2022-05-21
Add parquet.GenericBuffer, parquet.GenericReader, parquet.GenericWriter

achille-roussel:
This PR adds new generic APIs for programs compiling with Go 1.18+, which are equivalents to the existing `parquet.Buffer`, `parquet.Reader`, and `parquet.Writer`, but type-parameterized to have compile-time guarantees of the row types that are read or written.

These new APIs create many opportunities to create both safer and more efficient programs, which I tried to document in the source code.

I added benchmarks to compare the performance of the new APIs and worked on optimizing the writes.

```
goos: linux
goarch: amd64
pkg: github.com/segmentio/parquet-go
cpu: Intel(R) Xeon(R) CPU @ 2.60GHz
```

```
name           old time/op  new time/op   delta
GenericBuffer  15.6µs ± 2%    0.3µs ± 1%    -97.78%  (p=0.000 n=10+10)

name           old row/s    new row/s     delta
GenericBuffer   6.39M ± 2%  287.75M ± 1%  +4401.54%  (p=0.000 n=10+10)
```

```
name           old time/op  new time/op  delta
GenericWriter  17.6µs ± 2%   1.0µs ± 2%    -94.06%  (p=0.000 n=9+10)

name           old row/s    new row/s    delta
GenericWriter   5.68M ± 2%  95.75M ± 1%  +1587.10%  (p=0.000 n=9+9)
```

```
name           old time/op  new time/op  delta
GenericReader  9.87µs ± 3%  7.33µs ± 4%  -25.79%  (p=0.000 n=9+10)

name           old row/s    new row/s    delta
GenericReader   10.1M ± 3%   13.5M ± 4%  +33.53%  (p=0.000 n=9+10)
```

I will leave further optimization of the `parquet.Reader` type to a follow up.


buffer_go18.go
```diff
@@ -0,0 +1,176 @@
...
+// Go type representing the schema of rows in the buffer.
+//
+// See GenericWriter for details about the benefits over the classic Buffer API.
+type GenericBuffer[T any] struct {
+	base    Buffer
+	write   bufferFunc[T]
+	columns columnBufferWriter
+}
+
+// NewGenericBuffer is like NewBuffer but returns GenericBuffer[T] suited to write
```
kevinburkesegment, line 20:
```suggestion
// NewGenericBuffer is like NewBuffer but returns a GenericBuffer[T] suited to write
```

column_buffer_go18.go
```diff
@@ -0,0 +1,994 @@
...
+}
+
+// columnBufferWriter is a type used to writes rows to column buffers.
+type columnBufferWriter struct {
+	columns []ColumnBuffer // column buffers where row values are written
+	values  []Value        // buffer used to temporarily hold values
+	maxLen  int            // max number of values written to the temp buffer
+}
+
+// writeRowsFunc is the type of functions
```
kevinburkesegment, line 255: Incomplete sentence?

column_buffer_go18.go
```diff
@@ -0,0 +1,994 @@
...
+		// We need to find zero values, which should be represented as nulls
+		// in the parquet column. In order to minimize the calls to writeRows
+		// and maximize throughput, we use the nullIndex and nonNullIndex
+		// functions, which are type-specific implementations of the algorithm.
+		//
+		// Sections of the input that are contiguous nulls or non-nulls can be
+		// sent to a single call to writeRows to be written to the underlying
+		// buffer since they share the same definition level.
+		//
+		// This optimization is defeated by inputs alterning null and non-null
```
kevinburkesegment, line 366:
```suggestion
		// This optimization is defeated by inputs alternating null and non-null
```

column_buffer_go18.go
```diff
@@ -0,0 +1,994 @@
...
+			v.definitionLevel = levels.definitionLevel
+			w.values = append(w.values, v)
+		}
+		_, err = w.columns[levels.columnIndex].WriteValues(w.values)
+	}
+
+	return err
+}
+
+func (w *columnBufferWriter) writeRowsFloat32(rows array, size, offset uintptr, levels columnLevels) (err error) {
```
kevinburkesegment, line 834: Would be nice maybe to lean into code generation for some of these very repetitive functions.

internal/unsafecast/unsafecast.go
```diff
@@ -65,3 +65,12 @@ func BytesToSlice[T any](data []byte) []T {
 func BytesToString(data []byte) string {
 	return *(*string)(unsafe.Pointer(&data))
 }
+
+// StringToBytes applies the inverse conversion of BytesToString.
+func StringToBytes(data string) []byte {
```
kevinburkesegment, line 70: I'm so baffled that this is the recommended, best way to do this haha.

writer_go18.go
```diff
@@ -0,0 +1,211 @@
...
+
+// GenericWriter is similar to a Writer but uses a type parameter to define the
+// Go type representing the schema of rows being written.
+//
+// Using this type over Writer has multiple advantages:
+//
+// - By leveraging type information, the Go compiler can provide greater
+//   guarantees that the code is correct. For example, the parquet.Writer.Write
+//   method accepts an argument of type interface{}, which delays type checking
+//   at the runtime. The parquet.GenericWriter[T].Write method ensures at
```
kevinburkesegment, line 18:
```suggestion
//   until runtime. The parquet.GenericWriter[T].Write method ensures at
```

writer_go18.go
```diff
@@ -0,0 +1,211 @@
...
+//   compile time that the values it receives will be of type T, reducing the
+//   risk of introducing bugs.
```
kevinburkesegment, line 20:
```suggestion
//   risk of introducing errors.
```

Nit, I don't like "bugs" which implies something don't have much control over like an insect flying into your room.

writer_go18.go
```diff
@@ -0,0 +1,211 @@
...
+//
+// - Since type information is known at compile time, the implementation of
+//   parquet.GenericWriter[T] can make safe assumptions, removing the need for
+//   runtime validation of how the parameters passed to its methods.
```
kevinburkesegment, line 24:
```suggestion
//   runtime validation of how the parameters are passed to its methods.
```

kevinburkesegment (approved):

column_buffer_go18.go
achille-roussel: Yep.

Generics aren't in a place yet where they can yield the returns I think, we'd still need a few versions (e.g. when the type switch has more than one case), and there's a risk of seeing a measurable impact on the compute footprint.

We could use `go:generate` to render the code from a template, but this would bring quite a bit of complexity to the development process.

Typing a few functions is a one-time cost, it takes 15-20 minutes to do and the result is clear to the reader, which is an important aspect to optimize for.

internal/unsafecast/unsafecast.go
achille-roussel: I think `[]byte(s)` would be the recommended way, but it requires making a copy of the content since the byte slice is a mutable value.

The intent of this function is to avoid the copy.

PR #188, opened 2022-05-23
Optimize byte stream split encoding

achille-roussel:
I was interested to learn about how to used the VPGATHER* and VPSCATTER* instructions in x86, so I used the BYTE_STREAM_SPLIT encoding as a test case.

Overall, the instructions bring measurable throughput improvements compared to the scalar versions of the algorithms, but they have high latency and don't seem to benefit from CPU pipelining. I tried to unroll the loops but it did not bring any throughput improvements. 

The gains are nevertheless interesting, the assembly code using scalar instructions was 20-30% faster than what the Go compiler generated, but using the vector instruction yields unlocked a lot more efficiencies:

```
name                             old time/op    new time/op     delta
Encode/BYTE_STREAM_SPLIT/float     13.7µs ± 0%      4.3µs ± 0%   -68.41%  (p=0.000 n=10+9)
Encode/BYTE_STREAM_SPLIT/double    35.7µs ± 0%      5.5µs ± 1%   -84.63%  (p=0.000 n=9+10)
Decode/BYTE_STREAM_SPLIT/float     12.8µs ± 0%      2.1µs ± 0%   -83.65%  (p=0.000 n=8+8)
Decode/BYTE_STREAM_SPLIT/double    28.0µs ± 0%      3.4µs ± 1%   -87.95%  (p=0.000 n=8+10)

name                             old speed      new speed       delta
Encode/BYTE_STREAM_SPLIT/float   2.91GB/s ± 0%   9.22GB/s ± 0%  +216.54%  (p=0.000 n=10+9)
Encode/BYTE_STREAM_SPLIT/double  2.24GB/s ± 0%  14.57GB/s ± 1%  +550.60%  (p=0.000 n=9+10)
Decode/BYTE_STREAM_SPLIT/float   3.13GB/s ± 0%  19.15GB/s ± 0%  +511.38%  (p=0.000 n=8+10)
Decode/BYTE_STREAM_SPLIT/double  2.86GB/s ± 0%  23.73GB/s ± 1%  +730.11%  (p=0.000 n=8+10)

name                             old value/s    new value/s     delta
Encode/BYTE_STREAM_SPLIT/float       728M ± 0%      2305M ± 0%  +216.54%  (p=0.000 n=10+9)
Encode/BYTE_STREAM_SPLIT/double      280M ± 0%      1822M ± 1%  +550.60%  (p=0.000 n=9+10)
Decode/BYTE_STREAM_SPLIT/float       783M ± 0%      4789M ± 0%  +511.37%  (p=0.000 n=8+10)
Decode/BYTE_STREAM_SPLIT/double      357M ± 0%      2966M ± 1%  +730.09%  (p=0.000 n=8+10)
```


encoding/bytestreamsplit/bytestreamsplit_amd64.s
```diff
@@ -0,0 +1,376 @@
...
+    MOVQ R12, (AX)
+
+    ADDQ $8, AX
+    INCQ DX
+    CMPQ AX, CX
+    JB loop1x8
+done:
+    RET
+
+GLOBL scale8x4<>(SB), RODATA|NOPTR, $32
```
Pryz, line 328: Can you remind me what the following instructions do ?

achille-roussel: This is how you declare global variables in the Go assembly.

kevinburkesegment: Should we just leave a comment in the code explaining that? IMO the more assembler comments the better, especially if this code will end up being maintained by people who aren't as proficient with assembler

encoding/bytestreamsplit/bytestreamsplit_amd64.s
```diff
@@ -0,0 +1,376 @@
...
+DATA scale8x8<>+8(SB)/8, $1
+DATA scale8x8<>+16(SB)/8, $2
+DATA scale8x8<>+24(SB)/8, $3
+DATA scale8x8<>+32(SB)/8, $4
+DATA scale8x8<>+40(SB)/8, $5
+DATA scale8x8<>+48(SB)/8, $6
+DATA scale8x8<>+56(SB)/8, $7
+
+GLOBL shuffle8x8<>(SB), RODATA|NOPTR, $64
+DATA shuffle8x8<>+0(SB)/8,  $0x3830282018100800
```
kevinburkesegment, line 369: Also what are these constants?

kevinburkesegment: Like I am reading the pure go `encodeDouble` implementation and I don't see the relation between that and this

kevinburkesegment (approved):

encoding/bytestreamsplit/bytestreamsplit_amd64.s
achille-roussel: These are the shuffle masks for VPSHUFB/VPERMB:
* https://www.felixcloutier.com/x86/pshufb
* https://www.felixcloutier.com/x86/vpermb

The masks contain the indexes of bytes of the input that should be written at each index in the output.

I wrote them as hexadecimal numbers because that's the simplest way to declare the values of a byte array in the Go assembly, it's equivalent to the following Go code:
```go
var indexes = [...]byte{
  0: 0,
  1: 8,
  2: 16,
  3: 24,
  ...
}
```

encoding/bytestreamsplit/bytestreamsplit_amd64.s
achille-roussel: This is documented here https://go.dev/doc/asm

I don't think I want to document the assembler syntax, we can refer to the Go documentation for this. I left documentation explaining how the algorithms work in the last commit, let me know if you want more docs.

PR #190, opened 2022-05-23
fix slicing of repeated pages

achille-roussel:
This PR adds a test provided by @mdisibio which highlighted a bug, and fixes it.

The problem was a misuse of the repetition levels to detect the start of new rows. New rows start where the repetition level is zero, we were testing for `def != page.maxRepetitionLevel` instead of `def == 0`.

### DIFF
--- a/page.go
+++ b/page.go
@@ -444,27 +444,35 @@ func (page *repeatedPage) Slice(i, j int64) BufferedPage {
 		panic(errPageBoundsOutOfRange(i, j, numRows))
 	}
 
-	rowIndex0 := int64(0)
-	rowIndex1 := int64(len(page.repetitionLevels))
-	rowIndex2 := int64(len(page.repetitionLevels))
+	rowIndex0 := 0
+	rowIndex1 := len(page.repetitionLevels)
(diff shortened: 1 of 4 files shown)


page.go
```diff
@@ -444,27 +444,35 @@ func (page *repeatedPage) Slice(i, j int64) BufferedPage {
 		panic(errPageBoundsOutOfRange(i, j, numRows))
 	}
 
-	rowIndex0 := int64(0)
-	rowIndex1 := int64(len(page.repetitionLevels))
-	rowIndex2 := int64(len(page.repetitionLevels))
+	rowIndex0 := 0
+	rowIndex1 := len(page.repetitionLevels)
+	rowIndex2 := len(page.repetitionLevels)
```
kevinburkesegment, line 449: My first thought is that this is a typo since you're assigning two things to the same value where you might expect the second to be definitionLevels or whatever instead. Not sure how to make this clearer.

kevinburkesegment (approved):

page.go
achille-roussel: The idea is to set the values to the maximum, which would be the value we would land on if `i = j = numValues`.

PR #191, opened 2022-05-24
add CPU feature checks for AVX2

### LINKED ISSUE #189: Fallback to non AVX instructions when not available
We recently ran into an issue (https://github.com/parca-dev/parca/issues/1062) where a user had an amd64 architecture CPU without AVX extensions. This caused a `SIGILL` (illegal instruction) when parquet-go attempts to use the `VPBROADCASTD` instruction:

https://github.com/segmentio/parquet-go/blob/ce6d4da96469907924d6f87b746ca9169b6604cc/bloom/filter_amd64.s#L110

I would suggest we use the [golang.org/x/sys/cpu](https://pkg.go.dev/golang.org/x/sys/cpu) package to detect if the SIMD instructions we want to use are available and fallback to pure go implementations when not.

The apache arrow project accomplishes this using global variables:

https://github.com/apache/arrow/blob/442b24b0b9cf11d45245564fa7448f1ca4931ae7/go/arrow/math/int64.go#L30

Which are set using init functions:

https://github.com/apache/arrow/blob/6240eae2189d938b4f98b8d8235146baed0246c5/go/arrow/math/math_amd64.go#L25-L33

I'm happy with any strategy that works, this is just one I've seen before.

achille-roussel:
Fixes #189 

The `bloom` package was making the assumption that AVX2 would be available, however it appears that some users are running programs that depend on parquet-go on intel CPUs which do not have AVX2.

The PR adds a check for the AVX2 CPU feature in the assembly functions for amd64. The cost of this check is measurable in the benchmarks we have for bloom filters:

```
name              old time/op    new time/op    delta
Fasthash            1.18ns ± 1%    1.19ns ± 1%     ~     (p=0.735 n=9+10)
BlockInsert         2.35ns ± 1%    2.25ns ± 1%   -4.39%  (p=0.000 n=9+10)
BlockCheck          2.10ns ± 2%    2.39ns ± 2%  +13.75%  (p=0.000 n=10+10)
FilterInsertBulk    16.9ns ± 2%    17.3ns ± 3%   +2.42%  (p=0.001 n=10+9)
FilterInsert        2.40ns ± 3%    2.77ns ±11%  +15.64%  (p=0.000 n=10+10)
FilterCheck         2.69ns ± 1%    2.97ns ± 1%  +10.67%  (p=0.000 n=10+8)

name              old speed      new speed      delta
Fasthash          27.0GB/s ± 1%  26.9GB/s ± 1%     ~     (p=0.661 n=9+10)
BlockInsert       13.6GB/s ± 1%  14.2GB/s ± 1%   +4.61%  (p=0.000 n=9+10)
BlockCheck        15.2GB/s ± 2%  13.4GB/s ± 2%  -12.10%  (p=0.000 n=10+10)
FilterInsertBulk  30.2GB/s ± 2%  29.5GB/s ± 3%   -2.36%  (p=0.001 n=10+9)
FilterInsert      13.3GB/s ± 3%  11.6GB/s ±10%  -13.33%  (p=0.000 n=10+10)
FilterCheck       11.9GB/s ± 1%  10.8GB/s ± 1%   -9.64%  (p=0.000 n=10+8)
```

The `Block*` and `FilterInsert` functions aren't very important, `FilterInsertBulk` and `FilterCheck` are the functions used by the parquet-go package.

* The `FilterInsertBulk` benchmark measures a 2% drop in throughput, this seems acceptable to me, especially considering we can amortize this further by increasing the number of values in the bulk insert to reduce this overhead
* The `FilterCheck` benchmark measures a ~10% drop in throughput, this function tests a single value hash in a bloom filter, so it is more sensitive to the new instructions. The check runs in <3ns, so it seems to be an acceptable as well.

It appears to have no measurable impact on the high level `SplitBloomFilter` benchmark:
```
name              old time/op    new time/op    delta
SplitBlockFilter    1.94µs ± 1%    1.92µs ± 0%  -0.72%  (p=0.001 n=10+8)

name              old speed      new speed      delta
SplitBlockFilter  4.13GB/s ± 1%  4.16GB/s ± 0%  +0.73%  (p=0.001 n=10+8)
```

### DIFF
--- a/bloom/block_amd64.s
+++ b/bloom/block_amd64.s
@@ -2,6 +2,15 @@
 
 #include "textflag.h"
 
+#define salt0 0x47b6137b
+#define salt1 0x44974d91
+#define salt2 0x8824ad5b
+#define salt3 0xa2b7289d
+#define salt4 0x705495c7
+#define salt5 0x2df1424b
+#define salt6 0x9efc4947
+#define salt7 0x5c6bfb31
+
 DATA ones+0(SB)/4, $1
 DATA ones+4(SB)/4, $1
 DATA ones+8(SB)/4, $1
@@ -12,14 +21,14 @@ DATA ones+24(SB)/4, $1
 DATA ones+28(SB)/4, $1
 GLOBL ones(SB), RODATA|NOPTR, $32
 
-DATA salt+0(SB)/4, $0x47b6137b
-DATA salt+4(SB)/4, $0x44974d91
-DATA salt+8(SB)/4, $0x8824ad5b
-DATA salt+12(SB)/4, $0xa2b7289d
-DATA salt+16(SB)/4, $0x705495c7
-DATA salt+20(SB)/4, $0x2df1424b
-DATA salt+24(SB)/4, $0x9efc4947
-DATA salt+28(SB)/4, $0x5c6bfb31
+DATA salt+0(SB)/4, $salt0
+DATA salt+4(SB)/4, $salt1
+DATA salt+8(SB)/4, $salt2
+DATA salt+12(SB)/4, $salt3
+DATA salt+16(SB)/4, $salt4
+DATA salt+20(SB)/4, $salt5
+DATA salt+24(SB)/4, $salt6
+DATA salt+28(SB)/4, $salt7
 GLOBL salt(SB), RODATA|NOPTR, $32
 
 // This initial block is a SIMD implementation of the mask function declared in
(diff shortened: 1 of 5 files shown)


bloom/block_amd64.s
```diff
@@ -2,6 +2,15 @@
 
 #include "textflag.h"
 
+#define salt0 0x47b6137b
```
kevinburkesegment, line 5: I'm curious where these salt values come from.

kevinburkesegment:
I didn't realize we could just leave CPU instructions in the binary and as long as they don't get executed by the CPU that can't execute them, we're OK. That's pretty cool.

kevinburkesegment (approved):

bloom/block_amd64.s
achille-roussel: They come from the the parquet specs https://github.com/apache/parquet-format/blob/master/BloomFilter.md, I don't know how they generated them in the first place.

achille-roussel:
Thanks for the review!

PR #202, opened 2022-06-01
remove unused WriterConfig.PageBufferPool field

achille-roussel:


kevinburkesegment:
Would be great if we could get to a point where we could enable staticcheck on this repo, it can flag stuff like this.

PR #209, opened 2022-06-03
fix use of AVX-512 in DELTA_BINARY_PACKED encoding

achille-roussel:
VPMAXSQ is a AVX-512 instruction, since we were only gating for AVX2 the function would trigger SIGILL when running on hardware which did not support AVX-512.

### DIFF
--- a/encoding/delta/binary_packed_amd64.s
+++ b/encoding/delta/binary_packed_amd64.s
@@ -550,6 +550,11 @@ TEXT ·blockSubInt64AVX2(SB), NOSPLIT, $0-16
     VZEROUPPER
     RET
 
+// vpmaxsq is an emulation of the AVX-512 VPMAXSQ instruction with AVX2.
+#define vpmaxsq(tmp, arg2, arg1, ret) \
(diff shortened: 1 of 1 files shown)


encoding/delta/binary_packed_amd64.s
```diff
@@ -550,6 +550,11 @@ TEXT ·blockSubInt64AVX2(SB), NOSPLIT, $0-16
     VZEROUPPER
     RET
 
+// vpmaxsq is an emulation of the AVX-512 VPMAXSQ instruction with AVX2.
```
Pryz, line 553: Could it make sense to have two versions of this function to leverage AVX-512 when possible ? or would this add too much complexity for the gain ?

Pryz (approved):

encoding/delta/binary_packed_amd64.s
achille-roussel: We could, it doesn't always yield better perf tho, this would require further testing for this particular case.

PR #217, opened 2022-06-08
Optimize UUID types

achille-roussel:
The code was starting to be crippled with checks for fixed-length byte arrays of size 16 as a mechanism to use better code paths for the common case of using columns of type `[16]byte` (e.g. UUIDs, Trace IDs, etc...).

Rather than having these details spread through the codebase, I would like to introduce a new set of types which provide the internal specialization for these column types.

Having knowledege of the type size enables more optimizations as well, both on the Go side (see `compareBE128`, `lessBE128`), but also in assembly routines, for example in the dictionary encoding:

```
name                                        old time/op  new time/op  delta
Dictionary/Bounds/FIXED_LEN_BYTE_ARRAY(16)  8.05µs ± 0%  1.14µs ± 0%   -85.87%  (p=0.000 n=8+10)
Dictionary/Insert/FIXED_LEN_BYTE_ARRAY(16)  20.0µs ± 1%  20.9µs ± 1%    +4.32%  (p=0.000 n=10+9)
Dictionary/Lookup/FIXED_LEN_BYTE_ARRAY(16)   857ns ± 1%   856ns ± 0%      ~     (p=0.309 n=8+9)

name                                        old value/s  new value/s  delta
Dictionary/Bounds/FIXED_LEN_BYTE_ARRAY(16)    124M ± 0%    879M ± 0%  +607.77%  (p=0.000 n=8+10)
Dictionary/Insert/FIXED_LEN_BYTE_ARRAY(16)   50.0M ± 1%   47.9M ± 1%    -4.14%  (p=0.000 n=10+9)
Dictionary/Lookup/FIXED_LEN_BYTE_ARRAY(16)   1.17G ± 1%   1.17G ± 0%      ~     (p=0.236 n=8+9)
```

The increase on the insert path is due to the Go runtime optimizing the map implementation when the key is a `string`. The key type being changed to `[16]byte`, there is a small runtime cost. What the benchmark doesn't show tho is the fact that using `string` keys requires more heap allocations and memory copies.

### DIFF
--- a/dictionary_amd64.s
+++ b/dictionary_amd64.s
@@ -574,6 +574,79 @@ indexOutOfBounds:
     MOVQ $errnoIndexOutOfBounds, R12
     JMP return
 
+// func dictionaryBoundsBE128(dict [][16]byte, indexes []int32) (min, max *[16]byte, err errno)
+TEXT ·dictionaryBoundsBE128(SB), NOSPLIT, $0-72
+    MOVQ dict_base+0(FP), AX
+    MOVQ dict_len+8(FP), BX
+
+    MOVQ indexes_base+24(FP), CX
+    MOVQ indexes_len+32(FP), DX
+    SHLQ $2, DX // x 4
+    ADDQ CX, DX // end
+
+    XORQ R8, R8 // min (pointer)
+    XORQ R9, R9 // max (pointer)
+    XORQ SI, SI // err
+    XORQ DI, DI
+
+    CMPQ DX, $0
+    JE return
+
+    MOVL (CX), DI
+    CMPL DI, BX
+    JAE indexOutOfBounds
+    SHLQ $4, DI // the dictionary contains 16 byte words
+    LEAQ (AX)(DI*1), R8
+    MOVQ R8, R9
+    MOVBEQQ 0(AX)(DI*1), R10 // min (high)
+    MOVBEQQ 8(AX)(DI*1), R11 // min (low)
+    MOVQ R10, R12 // max (high)
+    MOVQ R11, R13 // max (low)
+
+    JMP next
+loop:
+    MOVL (CX), DI
+    CMPL DI, BX
+    JAE indexOutOfBounds
+    SHLQ $4, DI
+    MOVQ 0(AX)(DI*1), R14
+    MOVQ 8(AX)(DI*1), R15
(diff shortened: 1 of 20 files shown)


schema.go
```diff
@@ -612,6 +613,12 @@ func makeStructField(f reflect.StructField) structField {
 	return field
 }
 
+// FixedLenByteArray decimals are sized based on precision
+// this function calculates the necessary byte array size.
+func decimalFixedLenByteArraySize(precision int) int {
+	return int(math.Ceil((math.Log10(2) + float64(precision)) / math.Log10(256)))
+}
+
```
achille-roussel, line 621: I moved this function from `type.go` since it was only used here.

page_values.go
```diff
@@ -122,8 +122,7 @@ func (r *booleanPageValues) ReadBooleans(values []bool) (n int, err error) {
 
 func (r *booleanPageValues) ReadValues(values []Value) (n int, err error) {
 	for n < len(values) && r.offset < int(r.page.numValues) {
-		values[n] = makeValueBoolean(r.page.valueAt(r.offset))
-		values[n].columnIndex = r.page.columnIndex
+		values[n] = r.page.makeValue(r.page.valueAt(r.offset))
```
achille-roussel, line 125: I made a few changes to use the `makeValue` method of page types instead of repeating the assignment of column index here.

dictionary_amd64.s
```diff
@@ -814,6 +887,49 @@ indexOutOfBounds:
     MOVQ $errnoIndexOutOfBounds, AX
     JMP return
 
+// This is the same algorithm as dictionaryLookupFixedLenByteArrayString but we
+// only store the pointer to the location hodling the value instead of storing
```
kevinburkesegment, line 891:
```suggestion
// only store the pointer to the location holding the value instead of storing
```

Too much Bitcoin recently?

kevinburkesegment (approved):
Looks good to me, I would maybe make more explicit that the 128 bit BE type is for UUID's? It's clear in the PR but that relationship isn't clear in the code.

dictionary_amd64.s
achille-roussel: Ahah!

The internet will be the end of me!

Issue #221, opened 2022-06-08
Remove internal/bits package


achille-roussel:
The `internal/bits` package is lacking a clear purpose, it hosts a lot of code which has single call sites, and its name is conflicting with the `math/bits` package which sometimes makes maintenance more difficult than it could be.

This issue is open to track removal of the package, the suggested plan is:
- [x] move all type conversion functions to the `internal/unsafecast` package
https://github.com/segmentio/parquet-go/pull/222
- [x] move functions which single call sites to the package where they are used
https://github.com/segmentio/parquet-go/pull/224
- [x] remove functions when the abstraction does not add clear value to the code
https://github.com/segmentio/parquet-go/pull/223

PR #223, opened 2022-06-08
Internal bits remove unused code

achille-roussel:
This PR is a second step towards removing the `internal/bits` package https://github.com/segmentio/parquet-go/issues/221

The changes here remove unused code and unnecessary abstractions. See comments inline for details.


bloom/filter.go
```diff
@@ -47,9 +45,9 @@ func MakeSplitBlockFilter(data []byte) SplitBlockFilter {
 //	f := make(bloom.SplitBlockFilter, bloom.NumSplitBlocksOf(n, 10))
 //
 func NumSplitBlocksOf(numValues int64, bitsPerValue uint) int {
-	numBytes := bits.ByteCount(uint(numValues) * bitsPerValue)
+	numBytes := ((uint(numValues) * bitsPerValue) + 7) / 8
```
achille-roussel, line 48: This might be a part of the change that is more open to discussion, the `bits.ByteCount` function was used in more than one place.

I decided to inline it because the operation was really trivial and I could not come up with a good name for a package that would host this function.

encoding/encoding_test.go
```diff
@@ -904,3 +904,33 @@ func generateFixedLenByteArrayValues(n int, r *rand.Rand, size int) []byte {
...
+
+func maxLenInt64(data []int64) int {
+	max := 0
+	for _, v := range data {
+		if n := bits.Len64(uint64(v)); n > max {
+			max = n
+		}
+	}
+	return max
+}
```
achille-roussel, line 936: These functions were in the `internal/bits` package but only used here (with one exception), it seemed better to bring them here instead of having a dependency.

bloom/filter.go
kevinburkesegment: Fine by me

encoding/encoding_test.go
```diff
@@ -904,3 +904,33 @@ func generateFixedLenByteArrayValues(n int, r *rand.Rand, size int) []byte {
...
+	max := 0
+	for _, v := range data {
+		if n := bits.Len8(uint8(v)); n > max {
+			max = n
+		}
+	}
+	return max
+}
+
+func maxLenInt32(data []int32) int {
```
kevinburkesegment, line 918: I'm not super clear on what this returns, given the name. If it's just the largest item I'd expect "maxInt32". If it's the length you'd call "len", what am I missing?

kevinburkesegment (approved):

encoding/encoding_test.go
achille-roussel: "len" here is the length of the values in bits, as returned by `bits.Len*` functions in the standard `math/bits` package, I wanted to express the "maximum of all bits.Len in the list of int32".

Would you have a suggestion for a different name?

PR #225, opened 2022-06-09
Configure the capacity of column buffers as a number of values

achille-roussel:
The size of column buffers was configured as a number of bytes, this PR changes it to be expressed as a number of values, the intent being to improve usability of the library, and reduce memory waste.

Different column types have widely different memory footprints; considering a UUID column using 16 bytes per value, and a boolean column using 1 bit per value, there is a 128x ratio between the memory density of these columns. When the size of column buffers is expressed in bytes, we either under-allocate the buffers for columns of large types (resulting in more re-allocation and copies of the buffers), or over-allocate the buffers for columns of small types (resulting in a lot of unused memory). By expressing the capacity as a number of values, the library is able to better adjust the initial sizes of column buffers depending on the column types.

I believe that this change will also improve usability of the code: instead of having to reason in terms of how many bytes the application is expected to use, users can reason in terms of how many rows they are expecting to buffer in memory, which is a measure mapping more naturally to the application logic.

Please take a look and let me know if you have any comments or concerns about the change!

### DIFF
--- a/type.go
+++ b/type.go
@@ -62,6 +62,12 @@ type Type interface {
 	// For other types, the value is zero.
 	Length() int
 
+	// Returns an estimation of the number of bytes required to hold the given
+	// number of values of this type in memory.
+	//
+	// The method returns zero for group types.
+	EstimateSize(numValues int) int64
+
 	// Compares two values and returns a negative integer if a < b, positive if
 	// a > b, or zero if a == b.
 	//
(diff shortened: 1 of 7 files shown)


type.go
```diff
@@ -62,6 +62,12 @@ type Type interface {
 	// For other types, the value is zero.
 	Length() int
 
+	// Returns an estimation of the number of bytes required to hold the given
+	// number of values of this type in memory.
+	//
+	// The method returns zero for group types.
+	EstimateSize(numValues int) int64
```
kevinburkesegment, line 69: Can we be more precise about the estimate? Is it always an exact estimate OR overestimate, or is it always an exact estimate OR underestimate, or, neither

kevinburkesegment (approved):

type.go
achille-roussel: Sometimes it overestimates, sometimes it might underestimate, and sometimes it is an exact count.

For example, with boolean columns it always over-estimate since it rounds up to the nearest byte (the values are bit-packed).

For fixed-size types like int32, int64, etc... it's always the exact count.

For variable size types (BYTE_ARRAY) it's an approximation which may either underestimate or overestimate because we can't predict the length of values that the application will be working with.

I'm thinking there are risks with being more precise, users may try to be too smart for their own good by relying on the documented behavior, it also gives us less flexibility to change the behavior later on (if it is undocumented we have a stronger argument for "you should not have taken a dependency on this behavior").

I'm tempted to say that in this case we are better off making few promises, the value returned is a useful hint but users are not expected to rely on it beyond that.

What do you think?

kevinburkesegment: Yeah that's OK with me.

PR #227, opened 2022-06-09
Allow slice for FixedLengthByteArray decimals to improve performance

hhoughgg:
At the moment for fixed length byte array decimals the array is never addressable so it calls reflect.New for each value. There is some code to check the addressability of the array but my understanding is that it will never be addressable requiring this allocation each time. Using a slice of the correct size allows skipping this allocation.

For pointer fields currently the schema functions will show the kind as reflect.Pointer and cause it to throw an invalid field tag error. Added some code to check if the type is a pointer and get the underlying value with Elem.


achille-roussel (approved):
This change looks good to me, thanks for your contribution!

achille-roussel:
`reflect.Pointer` was introduced in Go 1.18, it used to be named `reflect.Ptr` which causes a build failure with Go 1.17:

```
# github.com/segmentio/parquet-go [github.com/segmentio/parquet-go.test]
Error: ./schema.go:441:22: undefined: reflect.Pointer
```

hhoughgg:
I had to make more changes as doing f.Type == f.Type.Elem caused other parts of the code to break since they do this check as well. Not sure if this is really the best way to do it but does scope the type to this single function.

achille-roussel (approved):
The change is looking good, thanks for the improvement!

Issue #226, opened 2022-06-09
Improve throughput of file merges


achille-roussel:
Merging parquet files is a critical operation for the use we are making of this library. Merging files is heavy on both the read and write paths, as files need to be decoded then re-encoded with the new content (e.g. rows may be re-ordered).

While we have invested in improving the write path, the algorithms involved on the read path are still trivial, yielding throughput way below the level of performance we need to achieve for production workloads.

This issue intends to track progress on improvement of the read throughput, especially in the context of merging parquet files.

annanay25:
+1 improving performance on the read path will be very useful for us as well. From some profiles we can see that ~60-70% of CPU on our compactors is spent in reading + reconstructing rows.

achille-roussel:
Progress so far:
```
name                                                 old time/op  new time/op  delta
MergeFiles/BOOLEAN/groups=2,rows=20000               17.0µs ± 1%  16.8µs ± 1%    -0.87%  (p=0.001 n=8+10)
MergeFiles/INT32/groups=2,rows=20000                 57.5µs ± 0%   6.9µs ± 1%   -87.93%  (p=0.000 n=8+8)
MergeFiles/INT64/groups=2,rows=20000                 6.41µs ± 0%  6.43µs ± 1%      ~     (p=1.000 n=9+8)
MergeFiles/INT96/groups=2,rows=20000                 33.5µs ±13%  36.4µs ±26%      ~     (p=0.393 n=10+10)
MergeFiles/FLOAT/groups=2,rows=20000                 6.21µs ± 1%  6.29µs ± 1%    +1.23%  (p=0.000 n=9+10)
MergeFiles/DOUBLE/groups=2,rows=20000                6.61µs ± 1%  6.65µs ± 1%    +0.65%  (p=0.010 n=9+10)
MergeFiles/BYTE_ARRAY/groups=2,rows=20000            46.0µs ± 1%  18.0µs ± 0%   -60.95%  (p=0.000 n=10+8)
MergeFiles/FIXED_LEN_BYTE_ARRAY/groups=2,rows=20000  7.67µs ± 5%  7.81µs ± 5%      ~     (p=0.148 n=10+10)
MergeFiles/STRING/groups=2,rows=20000                20.4µs ± 2%  11.1µs ± 0%   -45.73%  (p=0.000 n=8+9)
[...]

PR #230, opened 2022-06-12
Optimize delta decoding (part 2)

achille-roussel:
Following up from https://github.com/segmentio/parquet-go/pull/229, this PR contributes to https://github.com/segmentio/parquet-go/issues/226 by modifying the decoding of mini blocks in delta encoding using the techniques described in https://arxiv.org/pdf/1209.2137v5.pdf

I found these parts of the paper to be the best explanation of how the bit unpacking algorithm works:

image

> Figure 8. One step of simultaneous decoding of four 5-bit integers that are stored in a horizontal layout (as
opposed to the vertical data layout of Fig. 6). These integers are copied to four 32-bit words using the shuffle
operation pshufb. The locations in source and target buffers are indicated by arrows. Curvy lines are used
to denote integers that cross byte boundaries in the source buffer. Hence, they are copied only partially. The
boldface zero values represent the bytes zeroed by the shuffle instruction. Note that some source bytes are
copied to multiple locations.

_Page 13_

The implementation was quite complex, I iterated over it multiple times, starting with simple specializations and discovering how to make the algorithm more generic over time. I left comments in the code to capture some of the learning I thought could be relevant to future readers.

The following snapshots of benchmarks illustrate the impact of the change:
```
name                              old time/op    new time/op    delta
Decode/DELTA_BINARY_PACKED/int32    19.7µs ± 0%     9.4µs ± 0%   -52.16%  (p=0.000 n=10+10)

name                              old speed      new speed      delta
Decode/DELTA_BINARY_PACKED/int32  2.03GB/s ± 0%  4.25GB/s ± 0%  +109.04%  (p=0.000 n=10+10)

name                              old value/s    new value/s    delta
Decode/DELTA_BINARY_PACKED/int32      509M ± 0%     1063M ± 0%  +109.04%  (p=0.000 n=10+10)
```
```
name                                  old time/op  new time/op  delta
MergeFiles/INT32/groups=3,rows=30000  8.27µs ± 1%  7.11µs ± 2%  -14.05%  (p=0.000 n=10+9)
MergeFiles/UUID/groups=3,rows=30000   20.2µs ± 1%  18.4µs ± 1%   -9.32%  (p=0.000 n=10+9)

name                                  old row/s    new row/s    delta
MergeFiles/INT32/groups=3,rows=30000    119M ± 1%    138M ± 2%  +16.35%  (p=0.000 n=10+9)
MergeFiles/UUID/groups=3,rows=30000    48.6M ± 1%   53.6M ± 1%  +10.28%  (p=0.000 n=10+9)
```

Please take a look and let me know if you think anything could be improved.


encoding/delta/binary_packed_decode_amd64.s
```diff
@@ -134,6 +135,304 @@ test:
...
+//   and "vertical". The "horizontal" version is the one that applies the best
+//   to the bit packing done in the Parquet delta encoding; however, it also
+//   differs in some ways, many compression techniques discussed in the paper
+//   are not implemented in the Parquet format.
+//
+// - The paper focuses on implementations based on SSE instructions, which
+//   describes how to use PMULLD to emulate the lack of variable bit shift
+//   for packed integers. Our version of the bit unpacking algorithms here
+//   uses AVX2 and can perform variable bit shifts using VPSRLVD, which yields
+//   better throughput since the instruction latency is a single CPU cycles,
```
kevinburkesegment, line 153:
```suggestion
//   better throughput since the instruction latency is a single CPU cycle,
```

encoding/delta/binary_packed_decode_amd64.s
```diff
@@ -134,6 +135,304 @@ test:
...
+//   vs 10 for VPMULLD.
+//
+// - The reference implementation at https://github.com/lemire/FastPFor/ uses
+//   specializations for each bit size, resulting in 32 unique functions.
+//   Our version here are more generic, we provide 3 declinaisons of the
```
kevinburkesegment, line 158:
```suggestion
//   Our version here is more generic, we provide 3 declarations (?) of the
```

kevinburkesegment: French to English dictionary also suggests "declensions", either fine with me.

encoding/delta/binary_packed_decode_amd64.s
```diff
@@ -134,6 +135,304 @@ test:
...
+//   algorithms for bit widths 1 to 16, 17 to 26, and 27 to 31 (unpacking 32
+//   bits values is a simple copy). In that regard, our implementation is
+//   somewhat an improvement over the reference, it uses less code and less
```
kevinburkesegment, line 161:
```suggestion
//   somewhat an improvement over the reference, since it uses less code and less
```

kevinburkesegment (approved):

encoding/delta/binary_packed_decode_amd64.s
achille-roussel: I think what I meant was "variations" 👍

PR #231, opened 2022-06-12
Optimize delta decoding (part 3)

achille-roussel:
Following up to , this PR contributes to https://github.com/segmentio/parquet-go/issues/226 by optimizing the decoding of DELTA_LENGTH_BYTE_ARRAY.

This encoding is the default used for variable length byte arrays (e.g. Go's `string` type), which makes it a very critical path since a lot of data are modeled as strings.

The main difficulty in improving these code path came from the fact that we often deal with short string values, so calls to `runtime.memmove` (via `copy`) end up having a high overhead since there are no opportunities to use SIMD instructions for the copy when the input is just a few bytes long.

In parquet, we always decode full pages, so I used an opportunistic copy strategy to improve the copy throughput:
* if we ensure that we have some padding at the end of the input and output buffers, we can always read fixed size chunks without being concerned about overflowing beyond the buffers limits
* with this assumption, we can make a second assumption that short strings will always fit within a handful of bytes
* the algorithm can always read chunks of 16 bytes from the input, and often end after the first loop iteration since most strings will fit within the first
* for longer strings, we can perform copies in even larger chunks (32 bytes seemed to show good results)

This strategy appeared to be effective on our tests, the following snapshots show the impact on benchmarks comparing throughput with the previous versions of the code, on the function itself, on the full decoding of pages, and on the merging of files containing a BYTE_ARRAY column:

```
name                               old time/op    new time/op     delta
DecodeLengthByteArray/maxLen=0       4.15µs ± 0%     4.58µs ± 0%   +10.19%  (p=0.000 n=9+8)
DecodeLengthByteArray/maxLen=10      4.88µs ± 0%     0.71µs ± 0%   -85.36%  (p=0.000 n=9+9)
DecodeLengthByteArray/maxLen=20      5.47µs ±27%     0.95µs ± 0%   -82.59%  (p=0.000 n=10+8)
DecodeLengthByteArray/maxLen=100     6.05µs ± 9%     2.55µs ± 0%   -57.92%  (p=0.000 n=9+9)
DecodeLengthByteArray/maxLen=1000    17.8µs ± 1%     15.1µs ± 0%   -15.35%  (p=0.000 n=10+10)

name                               old speed      new speed       delta
DecodeLengthByteArray/maxLen=0      963MB/s ± 0%    874MB/s ± 0%    -9.25%  (p=0.000 n=9+8)
DecodeLengthByteArray/maxLen=10    1.99GB/s ± 0%  13.60GB/s ± 0%  +583.19%  (p=0.000 n=9+9)
DecodeLengthByteArray/maxLen=20    2.76GB/s ±22%  15.59GB/s ± 0%  +465.63%  (p=0.000 n=10+8)
DecodeLengthByteArray/maxLen=100   9.01GB/s ± 8%  21.38GB/s ± 0%  +137.34%  (p=0.000 n=9+9)
DecodeLengthByteArray/maxLen=1000  28.8GB/s ± 1%   34.1GB/s ± 0%   +18.12%  (p=0.000 n=10+10)
```

```
name                                       old time/op    new time/op    delta
Decode/DELTA_LENGTH_BYTE_ARRAY/byte_array    76.5µs ± 1%    20.4µs ± 0%   -73.34%  (p=0.000 n=10+9)

name                                       old speed      new speed      delta
Decode/DELTA_LENGTH_BYTE_ARRAY/byte_array  1.89GB/s ± 1%  7.09GB/s ± 0%  +275.04%  (p=0.000 n=10+9)

name                                       old value/s    new value/s    delta
Decode/DELTA_LENGTH_BYTE_ARRAY/byte_array      131M ± 1%      491M ± 0%  +275.03%  (p=0.000 n=10+9)
```

```
name                                       old time/op  new time/op  delta
MergeFiles/BYTE_ARRAY/groups=3,rows=30000  25.3µs ± 2%  18.5µs ± 1%  -26.93%  (p=0.000 n=10+9)

name                                       old row/s    new row/s    delta
MergeFiles/BYTE_ARRAY/groups=3,rows=30000   38.9M ± 2%   53.2M ± 1%  +36.85%  (p=0.000 n=10+9)
```

I think this strategy should also be applicable to other decoding, but also encoding paths, which should yield great improvements to all reads and writes of string values.


achille-roussel:
I applied the same decoding strategy to the DELTA_BYTE_ARRAY encoding which yielded similar results:

```
name                                             old time/op    new time/op    delta
Decode/DELTA_BYTE_ARRAY/byte_array                  151µs ± 3%      68µs ± 0%   -54.84%  (p=0.000 n=10+10)
Decode/DELTA_BYTE_ARRAY/fixed_length_byte_array     104µs ± 1%      25µs ± 0%   -76.23%  (p=0.000 n=9+9)

name                                             old speed      new speed      delta
Decode/DELTA_BYTE_ARRAY/byte_array                960MB/s ± 3%  2187MB/s ± 0%  +127.78%  (p=0.000 n=10+10)
Decode/DELTA_BYTE_ARRAY/fixed_length_byte_array  1.54GB/s ± 1%  6.47GB/s ± 0%  +320.63%  (p=0.000 n=9+9)

name                                             old value/s    new value/s    delta
Decode/DELTA_BYTE_ARRAY/byte_array                  66.4M ± 3%    147.1M ± 0%  +121.37%  (p=0.000 n=10+10)
Decode/DELTA_BYTE_ARRAY/fixed_length_byte_array     96.2M ± 1%    404.6M ± 0%  +320.62%  (p=0.000 n=9+9)
```

encoding/delta/length_byte_array_amd64.go
```diff
@@ -0,0 +1,84 @@
...
+	src = src[:totalLength]
+	dst = resizeNoMemclr(dst, size+padding)
+
+	i := 0
+	j := 0
+	k := 0
+	n := 0
+
+	// To leverage the SEE optimized implementation of the function we must
+	// create enough padding at the end to prevent the opportunisitic reads
```
kevinburkesegment, line 58:
```suggestion
	// create enough padding at the end to prevent the opportunistic reads
```

encoding/delta/length_byte_array_amd64.go
```diff
@@ -0,0 +1,84 @@
...
+	// and writes from overflowing past the buffers limits.
```
kevinburkesegment, line 59:
```suggestion
	// and writes from overflowing past the buffer's limits.
```

encoding/delta/length_byte_array_amd64.s
```diff
@@ -0,0 +1,139 @@
...
+// the input and outputs, even in the event where a value is shorter than this.
+//
+// The pointers to the current positions for input and output pointers are
+// always adjusted by the right number of bytes so that the next writes
+// overwrite any extra bytes that were written in the previous iteration of the
+// copy loop.
+//
+// The throughput of this function is not as good as runtime.memmove for large
+// buffers, but it ends up being close to an order of magnitude higher for the
+// common case of working with short strings.
```
kevinburkesegment, line 94: very clear 👍

encoding/delta/length_byte_array_amd64.s
```diff
@@ -0,0 +1,139 @@
...
+    CMPQ DX, DI
+    JNE loop
+    MOVQ dst_base+0(FP), BX
+    SUBQ BX, AX
+    SUBQ $4, AX
+    MOVQ AX, ret+72(FP)
+    VZEROUPPER
+    RET
+copy:
+    // Values longer than 16 bytes enter this loop and move 32 bytes chunks
```
kevinburkesegment, line 130:
```suggestion
    // Values longer than 16 bytes enter this loop and move 32 byte chunks
```

kevinburkesegment (approved):

PR #240, opened 2022-06-15
Optimize delta decoding (part 5)

achille-roussel:
This PR contributes to #226 by adding optimizations for decoding of 64 bit delta-encoded values.

I left comments inline, and focused on bit-packing between sizes 1 to 32 where the algorithm is a bit simpler than when values span across more than 32 bits. In practice, our use cases for delta-encoding on 64 bit values to have deltas <= MaxInt32 (e.g. encoding nearby timestamps which microsecond precision), we can revisit later and add specializations for the upper sizes if we measure it being a bottleneck.

I took a slightly different approach than in #230, I left documentation in the code explaining why.

```
name                     old time/op    new time/op     delta
UnpackInt64/bitWidth=1      163ns ± 0%       23ns ± 0%   -86.01%  (p=0.000 n=10+9)
UnpackInt64/bitWidth=8      164ns ± 0%       22ns ± 0%   -86.30%  (p=0.000 n=10+9)
UnpackInt64/bitWidth=10     186ns ± 0%       23ns ± 2%   -87.52%  (p=0.000 n=10+10)
UnpackInt64/bitWidth=20     204ns ± 1%       23ns ± 0%   -88.65%  (p=0.000 n=10+10)
UnpackInt64/bitWidth=30     241ns ± 1%       23ns ± 0%   -90.29%  (p=0.000 n=10+10)
UnpackInt64/bitWidth=40     252ns ± 1%      255ns ± 0%    +1.46%  (p=0.000 n=10+6)

name                     old speed      new speed       delta
UnpackInt64/bitWidth=1   3.14GB/s ± 0%  22.42GB/s ± 0%  +614.85%  (p=0.000 n=10+9)
UnpackInt64/bitWidth=8   3.13GB/s ± 0%  22.85GB/s ± 0%  +629.74%  (p=0.000 n=10+9)
UnpackInt64/bitWidth=10  2.76GB/s ± 0%  22.10GB/s ± 2%  +700.99%  (p=0.000 n=10+10)
UnpackInt64/bitWidth=20  2.51GB/s ± 1%  22.15GB/s ± 0%  +781.28%  (p=0.000 n=10+10)
UnpackInt64/bitWidth=30  2.13GB/s ± 1%  21.92GB/s ± 0%  +930.28%  (p=0.000 n=10+10)
UnpackInt64/bitWidth=40  2.03GB/s ± 1%   2.01GB/s ± 0%    -1.45%  (p=0.000 n=10+6)
```

```
name                                  old time/op  new time/op  delta
MergeFiles/INT64/groups=2,rows=20000  9.94µs ± 1%  7.48µs ± 1%  -24.76%  (p=0.000 n=9+9)

name                                  old row/s    new row/s    delta
MergeFiles/INT64/groups=2,rows=20000   97.4M ± 1%  129.4M ± 1%  +32.92%  (p=0.000 n=9+9)
```

### DIFF
--- a/internal/bitpack/unpack_int64_amd64.s
+++ b/internal/bitpack/unpack_int64_amd64.s
@@ -89,17 +89,16 @@ test:
 //
 // Since VPERMQ only supports reading the permutation combination from an
 // immediate value, we use VPERMD and generate permutation for pairs of two
-// consecutive 32 bit words, which is why we have set (x+1)<<32 to the upper
-// part of each 64 bit word.
+// consecutive 32 bit words, which is why we have the upper part of each 64
+// bit word set with (x+1)<<32.
 //
 // The masks for right shifts are written to Y5 and Y6, and computed with
 // this formula:
 //
 //      shift[i] = (bitWidth * i) - (32 * ((bitWidth * i) / 32))
 //
-// The amount to shift byte is the number of values previously unpacked,
-// offseted by the byte count of 32 bit words that we read from first bits
-// from.
+// The amount to shift by is the number of values previously unpacked, offseted
+// by the byte count of 32 bit words that we read from first bits from.
 //
 // Technically the masks could be precomputed and declared in global tables;
 // however, declaring masks for all bit width is tedious and makes code
--- a/internal/bitpack/unpack_int64_amd64.s
+++ b/internal/bitpack/unpack_int64_amd64.s
@@ -67,3 +67,119 @@ test:
     CMPQ SI, DX
     JNE loop
     RET
+
+// func unpackInt64x1to32bitsAVX2(dst []int64, src []byte, bitWidth uint)
+TEXT ·unpackInt64x1to32bitsAVX2(SB), NOSPLIT, $56-56
+    NO_LOCAL_POINTERS
+    MOVQ dst_base+0(FP), AX
+    MOVQ dst_len+8(FP), DX
+    MOVQ src_base+24(FP), BX
+    MOVQ bitWidth+48(FP), CX
+
+    CMPQ DX, $8
+    JB tail
+
+    MOVQ DX, DI
+    SHRQ $3, DI
+    SHLQ $3, DI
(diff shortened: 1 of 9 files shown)


internal/bitpack/unpack_int64_amd64.s
```diff
@@ -67,3 +67,134 @@ test:
...
+
+// This bit unpacking function was inspired from the 32 bit version, but
+// adapted to account for the fact that eight 64 bit values span across
+// two YMM registers, and across lanes of YMM registers.
+//
+// Because of the two lanes of YMM registers, we cannot use the VPSHUFB
+// instruction to dispatch bytes of the input to the registers. Instead we use
+// the VPERMD instruction, which has higher latency but supports dispatching
+// bytes across register lanes. Measurements show that throughput gains remain
+// very interesting despite the algorithm running on a few more CPU cycles per
```
kevinburkesegment, line 79:
```suggestion
// very interesting despite the algorithm running on a few more CPU cycles per
```

Better word than interesting? Gains remain persistent?

kevinburkesegment: Gains could also be interesting because they are *not* present, for example.

internal/bitpack/unpack_int64_amd64.s
```diff
@@ -67,3 +67,134 @@ test:
...
+//
+// The amount to shift byte is the number of values previously unpacked,
+// offseted by the byte count of 32 bit words that we read from first bits
+// from.
+//
+// Technically the masks could be precomputed and declared in global tables;
+// however, declaring masks for all bit width is tedious and makes code
+// maintenance more costly for no measurable benefits on production workloads.
+//
+// func unpackInt64x1to32bitsAVX2(dst []int64, src []byte, bitWidth uint)
```
kevinburkesegment, line 109: Thanks for writing this doc

kevinburkesegment (approved):

PR #242, opened 2022-06-16
Update documentation + add convenience APIs

achille-roussel:
This PR updates the README and code documentation to reflect the changes we introduced in #186, and add a few new APIs to simplify reading and writing parquet files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,7 +35,8 @@ dependency on and install with the following command:
 go get github.com/segmentio/parquet-go
 ```
 
-Go 1.17 or later is required to use the package.
+Go 1.17 is required to use the package, but some APIs and optimizations are only
+supported when building with Go 1.18 or later.
 
 ### Compatibility Guarantees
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -35,7 +35,8 @@ dependency on and install with the following command:
 go get github.com/segmentio/parquet-go
 ```
 
-Go 1.17 or later is required to use the package.
+Go 1.17 is required to use the package, but some APIs and optimizations are only
```
kevinburkesegment, line 38:
```suggestion
Go 1.17 is the lowest supported version. Some APIs and optimizations are only
```

kevinburkesegment: Previously you might think you _had_ to use 1.17 - saying it's "required" doesn't make sense in context if you are using 1.18

README.md
```diff
@@ -212,48 +232,50 @@ At this time, conversion rules only supports adding or removing columns from
 the schemas, there are no type conversions performed, nor ways to rename
 columns, etc... More advanced conversion rules may be added in the future.
 
-### Sorting Row Groups: [parquet.Buffer](https://pkg.go.dev/github.com/segmentio/parquet-go#Buffer)
+### Sorting Row Groups: [parquet.GenericBuffer[T]](https://pkg.go.dev/github.com/segmentio/parquet-go#Buffer)
 
-The `parquet.Writer` type is optimized for minimal memory usage, keeping the
-order rows unchanged and flushing pages as soon as they are filled.
+The `parquet.GenericWriter[T]` type is optimized for minimal memory usage,
+keeping the order rows unchanged and flushing pages as soon as they are filled.
```
kevinburkesegment, line 238:
```suggestion
keeping the order of rows unchanged and flushing pages as soon as they are filled.
```

README.md
```diff
@@ -212,48 +232,50 @@ At this time, conversion rules only supports adding or removing columns from
...
 calling their `Reset` method.
 
-The following example shows how to use a `parquet.Buffer` to order rows written
-to a parquet file:
+The following example shows how to use a `parquet.GenericBuffer[T]` to order rows
+written to a parquet file:
 
 ```go
-buffer := parquet.NewBuffer(
+buffer := parquet.NewGenericBuffer[RowType](
```
kevinburkesegment, line 263: Is the user supposed to define RowType? 

If so I'd maybe define it as part of the example - it's better to be explicit than implicit for docs, and to show the relationship between the objects.

README.md
```diff
@@ -446,23 +468,18 @@ The following examples demonstrate how to use these two models to write columns
...
-    values := make([]parquet.Value, len(columns[0]))
-    for i := range columns {
-        c := buffer.ColumnBuffers()[i]
-        for j, v := range columns[i] {
-            values[j] = parquet.ValueOf(v)
-        }
-        if _, err := c.WriteValues(values); err != nil {
-            return err
-        }
+func writeColumns(buffer *parquet.GenericBuffer[RowType], firstNames []string) error {
```
kevinburkesegment, line 471: Same comment here re: RowType

parquet_go18.go
```diff
@@ -0,0 +1,87 @@
+//go:build go1.18
+
+package parquet
+
+import (
+	"io"
+	"os"
+)
+
+// Read reads ands returns rows from the parquet file in the given reader.
```
kevinburkesegment, line 10:
```suggestion
// Read reads and returns rows from the parquet file in the given reader.
```

reader.go
```diff
@@ -29,6 +29,8 @@ import (
 //	}
 //
 //
+// For programs building with Go 1.18 or later, the GenericReader[T] type
+// supercedes this one.
```
kevinburkesegment, line 33:
```suggestion
// supersedes this one.
```

writer.go
```diff
@@ -41,6 +41,9 @@ import (
 // entire row group. Programs that need to produce parquet files with sorted
 // row groups should use the Buffer type to buffer and sort the rows prior to
 // writing them to a Writer.
+//
+// For programs building with Go 1.18 or later, the GenericWriter[T] type
+// supercedes this one.
```
kevinburkesegment, line 46:
```suggestion
// supersedes this one.
```

kevinburkesegment (approved):

README.md
achille-roussel: I can rephrase 👍

README.md
achille-roussel: Yes, `RowType` is whatever type the application wants to use since it's a generic.

Which do you think would be better:

- define `RowType` once in the first example
- define it every time (maybe fills the example with a lot of redundant and not very relevant info)
- use a more generic name like `[T]` of `[...]` to specify that this part of the code is not relevant to the example

achille-roussel:
Thanks for the feedback!

PR #246, opened 2022-06-21
[#189] do not use MOVBE instruction when compiling for x86

### LINKED ISSUE #189: Fallback to non AVX instructions when not available

achille-roussel:
It appears MOVBE was not supported on Intel Ivy Bridge CPUs. This PR replaces the use of MOVBE by the combination of MOV+BSWAP.

Fixes #189

### DIFF
--- a/page_min_amd64.s
+++ b/page_min_amd64.s
@@ -560,13 +560,17 @@ loop16:
     //
     // This loop is also take if the CPU has no support for AVX-512.
 loop:
-    MOVBEQQ (AX), R8
-    MOVBEQQ (BX), R9
+    MOVQ (AX), R8
+    MOVQ (BX), R9
+    BSWAPQ R8
+    BSWAPQ R9
     CMPQ R8, R9
     JB less
     JA next
-    MOVBEQQ 8(AX), R8
-    MOVBEQQ 8(BX), R9
+    MOVQ 8(AX), R8
+    MOVQ 8(BX), R9
+    BSWAPQ R8
+    BSWAPQ R9
     CMPQ R8, R9
     JAE next
 less:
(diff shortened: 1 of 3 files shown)


javierhonduco:
Thanks so much for the prompt fix! This addresses the problem I was experiencing 🎉

kevinburkesegment (approved):

page_min_amd64.s
```diff
@@ -560,13 +560,17 @@ loop16:
     //
     // This loop is also take if the CPU has no support for AVX-512.
```
kevinburkesegment, line 561:
```suggestion
    // This loop is also taken if the CPU has no support for AVX-512.
```

achille-roussel:
Thanks for the reviews!

PR #247, opened 2022-06-22
Select bitpack encoding and decoding implementation at runtime

javierhonduco:
👋🏻 Hi!! This PR adds runtime selection of the bitpack encoding and decoding implementation to addresses the last issue I am bumping into related to `SIGILL`s we were seeing while investigating https://github.com/segmentio/parquet-go/issues/189. 

On current main (`69748705dd0`):

```shell
$ go test . -c && gdb parquet-go.test
 (gdb) display/i $pc
1: x/i $pc
=> 0x717429 <[...]/encoding/rle.encodeBytesBitpack+41>:	pext   %rdi,%r8,%r8
```

The `PEXT` instructions seems to have been introduced with [Haswell](https://en.wikipedia.org/wiki/X86_Bit_manipulation_instruction_set).

Let me know why do you think, and thanks for all your work!! 😄 

## Test plan
```
[javierhonduco@computer parquet-go]$ cat /proc/cpuinfo | grep name | head -1
model name	: Intel(R) Core(TM) i5-3450 CPU @ 3.10GHz
[javierhonduco@computer parquet-go]$ go test -race .
ok  	github.com/segmentio/parquet-go	16.047s
```

### DIFF
--- a/encoding/rle/rle_amd64.go
+++ b/encoding/rle/rle_amd64.go
@@ -9,6 +9,8 @@ import (
 var (
 	encodeInt32IndexEqual8Contiguous func(words [][8]int32) int
 	encodeInt32Bitpack               func(dst []byte, src [][8]int32, bitWidth uint) int
+	encodeBytesBitpack               func(dst []byte, src []uint64, bitWidth uint) int
+	decodeBytesBitpack               func(dst, src []byte, count, bitWidth uint)
 )
 
 func init() {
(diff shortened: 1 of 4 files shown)


encoding/rle/rle_amd64.go
```diff
@@ -9,21 +9,29 @@ import (
...
 		encodeInt32Bitpack = encodeInt32BitpackDefault
+		encodeBytesBitpack = encodeBytesBitpackDefault
+		decodeBytesBitpack = decodeBytesBitpackDefault
+
 	}
 }
 
 //go:noescape
-func encodeBytesBitpack(dst []byte, src []uint64, bitWidth uint) int
+func encodeBytesBitpackASM(dst []byte, src []uint64, bitWidth uint) int
```
achille-roussel, line 34: Maybe named this `encodeBytesBitpackBMI2` since the instruction it relies on is part of the BMI2 instruction set? It would follow the pattern we used for other functions which are suffixed with AVX2 for example.

encoding/rle/rle_amd64.go
```diff
@@ -9,21 +9,29 @@ import (
...
+	decodeBytesBitpack               func(dst, src []byte, count, bitWidth uint)
 )
 
 func init() {
 	switch {
 	case cpu.X86.HasAVX2:
 		encodeInt32IndexEqual8Contiguous = encodeInt32IndexEqual8ContiguousAVX2
 		encodeInt32Bitpack = encodeInt32BitpackAVX2
+		encodeBytesBitpack = encodeBytesBitpackASM
+		decodeBytesBitpack = decodeBytesBitpackASM
```
achille-roussel, line 22: How about installing these if `cpu.X86.HasBMI2` is true? It would be more accurate than testing for AVX2.

encoding/rle/rle_amd64.go
javierhonduco: Good call!

encoding/rle/rle_amd64.go
javierhonduco: This is sweet! Completely missed it, thanks! 😄

achille-roussel (approved):
Thanks for the contribution!

Issue #156, opened 2022-04-25
Support field IDs in parquet schemas


achille-roussel:
The parquet format supports storing fields ids, which is especially useful to have when converting data between protobuf or thrift definitions.

ohkinozomu:
Hello, what is the goal of this issue?
I have created a PR https://github.com/segmentio/parquet-go/pull/253 , but it should be incomplete.

PR #253, opened 2022-06-24
Support FieldID

ohkinozomu:
Resolve https://github.com/segmentio/parquet-go/issues/156

### DIFF
--- a/column.go
+++ b/column.go
@@ -75,6 +75,9 @@ func (c *Column) Path() []string { return c.path }
 // Name returns the column name.
 func (c *Column) Name() string { return c.schema.Name }
 
+// FieldID returns the field ID.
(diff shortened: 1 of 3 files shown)


achille-roussel (commented):
Thanks for the contribution.

I believe the proposed solution here would only partially add support for field ids, am I correct?

We would be able to read them off of encoded parquet schemas (maybe generated by other libraries), but parquet-go would still offer no solutions allowing the application to define field ids.

ohkinozomu:
Yes, I didn't understand the goal of the Issue https://github.com/segmentio/parquet-go/issues/156 ...

> We would be able to read them off of encoded parquet schemas (maybe generated by other libraries), but parquet-go would still offer no solutions allowing the application to define field ids.

I have tried to work on it but need to get more familiar with the code.
I close this PR for now, thanks.

PR #255, opened 2022-06-30
Allocate columns to the exact size when reading file.

cyriltovena:
I decide to return the total when parsing columns to avoid extra traversal but I'm happy to switch to another traversal if that's what we prefer.

### DIFF
--- a/file.go
+++ b/file.go
@@ -85,7 +85,7 @@ func OpenFile(r io.ReaderAt, size int64, options ...FileOption) (*File, error) {
 	}
 
 	schema := NewSchema(f.root.Name(), f.root)
-	columns := make([]*Column, 0, MaxColumnIndex+1)
+	columns := make([]*Column, 0, numLeafColumnsOf(f.root))
 	f.schema = schema
 	f.root.forEachLeaf(func(c *Column) { columns = append(columns, c) })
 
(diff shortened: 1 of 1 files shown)


file.go
```diff
@@ -79,13 +79,14 @@ func OpenFile(r io.ReaderAt, size int64, options ...FileOption) (*File, error) {
...
-	if f.root, err = openColumns(f); err != nil {
+	var totalColumns int
+	f.root, totalColumns, err = openColumns(f)
+	if err != nil {
 		return nil, fmt.Errorf("opening columns of parquet file: %w", err)
 	}
 
 	schema := NewSchema(f.root.Name(), f.root)
-	columns := make([]*Column, 0, MaxColumnIndex+1)
+	columns := make([]*Column, 0, totalColumns)
```
achille-roussel, line 89: Maybe an alternative approach could be the following:
```suggestion
	columns := make([]*Column, 0, numLeafColumnsOf(f.root))
```

achille-roussel (commented):
The code looks good overall, let me know what you think of the comment I left!

file.go
cyriltovena: Yeah I though about going through it again, it has the advantages of clarity but opted for the other option to avoid recursion and another trip down the tree. I figured with complex schema it would perform better if you open a lot of file.

I'm not super strong on that opinion though so if you prefer clarity, I'll do the change.

achille-roussel: I would usually default to simplicity unless we have been able to measure the compute cost. I agree that intuitively the second pass might appear as "more work" but computers are not always behaving in intuitive ways (e.g. when the data resides in CPU caches the extra pass is likely to be just as efficient as doing it in one run).

cyriltovena: Sounds good will make the change

PR #258, opened 2022-07-06
Expose parquet file metadata.

cyriltovena:
I wanted to write an inspect tools to see the size of each columns but the metadata was hidden, this allows to deeply inspect parquet file and understand which columns are big.


achille-roussel (approved):
Thanks for the contribution!

PR #260, opened 2022-07-06
Optimize dictionaries for fixed-length types

achille-roussel:
This PR contributes to https://github.com/segmentio/parquet-go/issues/226 by adding a new `hashprobe` package which provides optimized implementations of probing tables, and using these data structures as replacement for the use of Go maps to back parquet dictionaries.

The gains come from two key observations:
* the Go map API does not offer the ability to perform bulk operations, resulting in spending a lot of CPU time calling into the runtime map functions
* the Go map implementation has to make performance trade offs to support operations like deletion and iteration, which we do not need in parquet dictionary, creating the opportunity to use a more efficient memory layout

I also added two subpackages: `hashprobe/aeshash` and `hashprobe/wyhash` providing implementations of hash functions similar to the ones used by the Go runtime.

The change measurable improvements compared to the use of Go maps:
```
name                                        old time/op  new time/op  delta
Dictionary/Insert/INT32                     7.37µs ± 0%  2.47µs ± 0%   -66.49%  (p=0.000 n=9+9)
Dictionary/Insert/INT64                     7.48µs ± 1%  2.69µs ± 0%   -64.00%  (p=0.000 n=10+10)
Dictionary/Insert/FLOAT                     21.1µs ± 1%   2.4µs ± 0%   -88.52%  (p=0.000 n=10+9)
Dictionary/Insert/DOUBLE                    22.7µs ± 0%   2.8µs ± 1%   -87.57%  (p=0.000 n=10+10)
Dictionary/Insert/FIXED_LEN_BYTE_ARRAY(16)  21.5µs ± 1%   4.9µs ± 0%   -77.19%  (p=0.000 n=10+10)
Dictionary/Insert/INT(32,false)             7.49µs ± 1%  2.52µs ± 0%   -66.39%  (p=0.000 n=10+10)
Dictionary/Insert/INT(64,false)             7.43µs ± 1%  2.76µs ± 1%   -62.83%  (p=0.000 n=10+10)

name                                        old value/s  new value/s  delta
Dictionary/Insert/INT32                       136M ± 0%    405M ± 0%  +198.33%  (p=0.000 n=9+10)
Dictionary/Insert/INT64                       134M ± 1%    371M ± 0%  +177.77%  (p=0.000 n=10+10)
Dictionary/Insert/FLOAT                      47.5M ± 1%  413.4M ± 0%  +770.91%  (p=0.000 n=10+9)
Dictionary/Insert/DOUBLE                     44.1M ± 0%  354.5M ± 1%  +704.21%  (p=0.000 n=10+10)
Dictionary/Insert/FIXED_LEN_BYTE_ARRAY(16)   46.4M ± 1%  203.5M ± 0%  +338.37%  (p=0.000 n=10+10)
Dictionary/Insert/INT(32,false)               133M ± 1%    397M ± 0%  +197.51%  (p=0.000 n=10+10)
Dictionary/Insert/INT(64,false)               135M ± 1%    363M ± 0%  +169.27%  (p=0.000 n=10+9)
```

I have a few follow up but I think this PR has enough content to be reviewed already:
* validation of `parquet.Value` inserted to BE128 dictionaries is important but costs over 40% of the compute footprint, this looks like a good are for improvement considering this type would be often used for dictionaries of UUID-like values
* I have not yet investigated dictionaries of variable-length values (e.g. `string`), but these are very popular models in parquet so worth providing a specialization for


column_buffer_amd64.go
```diff
@@ -33,7 +33,7 @@ func writeValuesBitpackAVX2(values unsafe.Pointer, rows array, size, offset uint
 //go:noescape
 func writeValues32bitsAVX2(values unsafe.Pointer, rows array, size, offset uintptr)
 
-//go:noescpae
+//go:noescape
```
achille-roussel, line 36: This was a bug which caused local variables to escape to the heap since the compiler could not determine whether the function would retain pointers that were passed as arguments.

It was surprising that neither `go build` nor `go vet` picked it up 🤷 I was hoping they would warn about invalid compiler directives.

parquet_go18.go
```diff
@@ -29,7 +29,7 @@ func Read[T any](r io.ReaderAt, size int64, options ...ReaderOption) (rows []T,
 	reader := NewGenericReader[T](file, config)
 	n, err := reader.Read(rows)
 	if n < len(rows) && err == io.EOF {
-		rows, err = rows[:n], err
+		rows, err = rows[:n], nil
```
achille-roussel, line 32: This bug was picked up by `go vet`

array.go
```diff
@@ -33,15 +33,15 @@ func (a array) index(i int, size, offset uintptr) unsafe.Pointer {
 	return unsafe.Add(a.ptr, uintptr(i)*size+offset)
 }
 
-func (a array) slice(i, j int, size, offset uintptr) array {
+func (a array) slice(i, j int, size uintptr) array {
```
achille-roussel, line 36: This was a bug, when slicing we need the subarray to start at the beginning of an element, not at the offset of a field.

column_buffer_amd64.go
```diff
@@ -33,7 +33,7 @@ func writeValuesBitpackAVX2(values unsafe.Pointer, rows array, size, offset uint
 //go:noescape
 func writeValues32bitsAVX2(values unsafe.Pointer, rows array, size, offset uintptr)
 
-//go:noescpae
```
kevinburke, line 36: Ugh, yikes. I wonder if a linter would have caught this?

dictionary.go
```diff
@@ -8,14 +8,33 @@ import (
...
-	// more representative of the map implementation in Go.
-	mapSizeOverheadPerItem = 8
+	// Maximum load of probing tables. This parameter configures the balance
+	// between memory density and compute time of probing operations.
+	//
+	// Smaller values result in lower collision probability when inserting
+	// values in probing tables, but also increase memory utilization.
+	//
+	// TODO: make this configurable by the application?
+	hashprobeTableMaxLoad = 0.9
```
kevinburkesegment, line 24: What is the range of acceptable values here? (0, 1) ?

hashprobe/hashprobe.go
```diff
@@ -0,0 +1,643 @@
...
+//		...
+//	}
+//
+// The method takes an array of keys to probe as first argument, an array of
+// values where the indexes of each key will be written as second argument, and
+// returns the number of keys that were inserted during the call.
+//
+// Applications that need to determine which keys were inserted can capture the
+// length of the probing table prior to the call, and scan the list of values
+// looking for indexes greater or equal to the captured lenth.
```
kevinburkesegment, line 24:
```suggestion
// looking for indexes greater or equal to the captured length.
```

I'm not super sure what this means. Maybe you could add a helper method or an example?

hashprobe/hashprobe.go
```diff
@@ -0,0 +1,643 @@
...
+	// stack but mostly erasing the baseline cost.
+	probesPerLoop = 256
+)
+
+func nextPowerOf2(n int) int {
+	return 1 << (64 - bits.LeadingZeros64(uint64(n-1)))
+}
+
+func randSeed() uintptr {
+	return uintptr(rand.Uint64())
```
kevinburkesegment, line 56: Don't you need to seed rand first? You'll always get the same value otherwise

kevinburkesegment (approved):

hashprobe/aeshash/aeshash_amd64.s
```diff
@@ -0,0 +1,150 @@
...
+    AESENC X1, X0
+    AESENC X2, X0
+    AESENC X3, X0
+
+    MOVQ X0, ret+16(FP)
+    RET
+
+// func Hash64(value uint64, seed uintptr) uintptr
+TEXT ·Hash64(SB), NOSPLIT, $0-24
+    MOVL value+0(FP), AX
```
chriso, line 26: Should it be `MOVQ` here and `MOVL` when loading the `value` in `Hash32` above?

hashprobe/hashprobe.go
```diff
@@ -0,0 +1,643 @@
...
+//
+// - The array of keys is written right after the bitmap. All keys are 32 bits
+//   values.
+//
+// - Values are written after the keys, with each slot of the keys array mapping
+//   to a value at the same slot in the values array.
+//
+// This memory layout is a combination between simplicity and performance; the
+// bitmap is required in order to differentiate between key slots with the zero
+// value, and key slots that have not yet been assigned.
```
chriso, line 125: An alternative would to reserve the zero valued hash? You could update the hashing routines to rehash with a different (but derived) seed if the hash result is zero. A zero should be exceedingly rare, and a branch not taken in the hashing routines is essentially free. This would let you avoid the bitmaps entirely.

hashprobe/hashprobe.go
```diff
@@ -0,0 +1,643 @@
...
+//
+// The memory layout is also optimized for fast probing of existing keys, in
+// which case the lookup can be performed with only 4 memory loads, and 1 store:
+// 1. load the bitmap location and test the bit to see that the slot is occupied
+// 2. load the key at the corresponding index, compare with the current key
+// 3. load the value at the corresponding index, save in the output buffer
+//
+// Conflicts are resolved by testing the next slot, until a free slot is found.
```
chriso, line 133: aka. https://en.wikipedia.org/wiki/Linear_probing

hashprobe/hashprobe.go
```diff
@@ -0,0 +1,643 @@
...
+	hashes := make([]uintptr, len(keys), 32)
+	modulo := uintptr(t.cap) - 1
+
+	if aeshash.Enabled() {
+		aeshash.MultiHash32(hashes, keys, t.seed)
+	} else {
+		wyhash.MultiHash32(hashes, keys, t.seed)
+	}
+
+	for i, hash := range hashes {
```
chriso, line 215: Would it make sense to use the multiprobe routines when growing the hash table?

chriso (commented):
This is awesome :fire:

hashprobe/hashprobe.go
achille-roussel: I'll try it out.

hashprobe/hashprobe.go
achille-roussel: I'll rephrase.

hashprobe/hashprobe.go
achille-roussel: I can improve this, I was thinking the application usually seeds `math/rand`, it would be weird for the parquet-go library to force seeding the global random number generator; the current form does seem like a potential landmine.

dictionary.go
achille-roussel: Yes

hashprobe/aeshash/aeshash_amd64.s
achille-roussel:
```
$ go vet ./hashprobe/...
# github.com/segmentio/parquet-go/hashprobe/aeshash
hashprobe/aeshash/aeshash_amd64.s:7:1: [amd64] Hash32: invalid MOVQ of value+0(FP); uint32 is 4-byte value
hashprobe/aeshash/aeshash_amd64.s:26:1: [amd64] Hash64: invalid MOVL of value+0(FP); uint64 is 8-byte value
```
👍

hashprobe/hashprobe.go
achille-roussel: I started an experiment in https://github.com/segmentio/parquet-go/tree/hashprobe-no-zero-hash tho it appears that testing if the hash value is zero incurs a ~10% drop in throughput on dictionary operations:
```
name                     old time/op  new time/op  delta
Dictionary/Insert/INT32  2.55µs ± 0%  2.83µs ± 0%  +11.31%  (p=0.000 n=9+10)

name                     old value/s  new value/s  delta
Dictionary/Insert/INT32    393M ± 0%    353M ± 0%  -10.16%  (p=0.000 n=9+10)
```
It's possible that there is a balance between reducing the number of instructions in the probing routines which results in better throughput overall, the extra instructions on the hashing code paths seem expensive tho.

achille-roussel: I don't know if the branch is the problem, it might just be the extra instructions that add the cost.

AES hash for example does 3 `AESENC` instructions per loop which are 5 cycles each (15 cycles per loop plus a few for the loop increment and test).

To do a `TEST` or `CMP` we need to extract the value from a vector to a general purpose register, overall it may just cost too many CPU cycles in the hashing function to be worth removing cycles in the probing function.

I'm also a bit concerned about modifying the hashing algorithm; if we diverge, we cannot test against the Go runtime functions to ensure that our implementation is correct, and we might degrade the quality of the hash functions in ways that are hard to predict.

chriso: There is a VPTEST instruction you can use to avoid the vector=>GP reg transfer cost, but yeah it might just be a case of too many extra instructions on the critical path.

PR #263, opened 2022-07-11
Do not copy values to temporary memory buffer when inserting values into dictionaries

achille-roussel:
Based on #262, this PR removes a measurable overhead of copying values to insert into dictionaries to a temporary buffer. I had doing this because the values may have come from sparse memory locations (e.g. packed in a `parquet.Value`) and needed to be presented as a contiguous array of keys to the probing tables.

I introduced a `hashprobe/sparse` package which offers an abstraction layer for accessing arrays of values in sparse memory locations, and allows the underlying algorithms to work on non-contiguous memory areas. I think this package could be reused in other places in parquet-go, I'll see how I could refactor the code to leverage it elsewhere in a follow up.

I measured a significant performance regression on large dictionary inserts (1M+ values), which I tracked down to being caused by CPU cache invalidation causing multiple passes over the memory areas to amplify the cost of cache misses. I figured out a mechanism for chunking the inserts so that the second passes are able to hit CPU caches. I documented these learnings in the code as well.

```
name                                                  old time/op  new time/op  delta
Dictionary/Insert/INT32/N=100                          237ns ± 0%   227ns ± 1%   -4.32%  (p=0.000 n=10+10)
Dictionary/Insert/INT32/N=1000                        2.15µs ± 0%  1.98µs ± 0%   -7.78%  (p=0.000 n=10+9)
Dictionary/Insert/INT32/N=10000                       24.3µs ± 0%  22.2µs ± 0%   -8.50%  (p=0.000 n=10+9)
Dictionary/Insert/INT32/N=100000                       361µs ± 6%   328µs ± 2%   -9.08%  (p=0.000 n=10+10)
Dictionary/Insert/INT32/N=1000000                     8.32ms ± 9%  8.02ms ± 4%     ~     (p=0.095 n=10+9)
Dictionary/Insert/INT64/N=100                          255ns ± 0%   216ns ± 0%  -15.26%  (p=0.000 n=10+10)
Dictionary/Insert/INT64/N=1000                        2.29µs ± 0%  2.11µs ± 0%   -7.66%  (p=0.000 n=10+10)
Dictionary/Insert/INT64/N=10000                       26.5µs ± 0%  23.8µs ± 0%  -10.24%  (p=0.000 n=9+10)
Dictionary/Insert/INT64/N=100000                       748µs ± 2%   708µs ± 2%   -5.27%  (p=0.000 n=10+9)
[...]

PR #264, opened 2022-07-11
Add sparse package

achille-roussel:
Follow up to #263, this PR moves all sparse array operations to a top level `sparse` package, and refactors parquet-go to use the package where it makes sense.

Apart from the new APIs of the `sparse` package, no functionalities were added here, this is only a refactor.


kevinburkesegment (approved):

sparse/sparse.go
```diff
@@ -0,0 +1,20 @@
+// Package sparse contains abstractions to help work on arrays of values in
+// sparse memory locations.
+//
+// Conversion between array types are supported when converting integers to a
```
kevinburkesegment, line 4:
```suggestion
// Conversion between array types is supported when converting integers to a
```

"Conversion is"

sparse/sparse.go
```diff
@@ -0,0 +1,20 @@
...
+// lower size (e.g. int32 to int16, or uint64 to uint8), or converting from
+// signed integers to unsigned. Float types can also be converted to unsigned
+// integers of the same size, in which case the conversion is similar to using
+// the standard library's math.Float32bits and math.Float64bits functions.
+//
+// All array types can be converted to a generic Array type can be used to erase
```
kevinburkesegment, line 10:
```suggestion
// All array types can be converted to a generic Array type that can be used to erase
```

sparse/gather_test.go
```diff
@@ -0,0 +1,140 @@
...
+	buf := make([]point2D, N+1)
+	dst := make([]uint64, N)
+	src := sparse.UnsafeUint64Array(unsafe.Pointer(&buf[0].Y), len(buf), unsafe.Sizeof(buf[0]))
+
+	for i := range buf {
+		buf[i].X = math.MaxUint64
+		buf[i].Y = uint64(2 * i)
+	}
+
+	if n := sparse.GatherUint64(dst, src); n != N {
```
kevinburkesegment, line 54: Add example_test.go with code examples here?

PR #267, opened 2022-07-12
Fix de-/reconstruction of pointer slices

simonswine:
This adds a test and a potential fix for destruction/reconstruction of a struct with a pointer slice.


achille-roussel (approved):
Thanks for the fix!

PR #268, opened 2022-07-12
Add the ability to pass Parquet tags on map key and values

### LINKED ISSUE #256: Add ability to set properties on key and value of a Parquet map
Currently when defining a Parquet map:

```
type A struct {
  B map[string]string `parquet:","`
}
```

We don't have the ability to set the properties of the key or value for that map.

One possibility is to add something similar to the following:

```
type A struct {
  B map[string]string `parquet:",key(dict,zstd)"`
}
```

Pryz:
Fixes #256

### DIFF
--- a/column_path.go
+++ b/column_path.go
@@ -1,6 +1,8 @@
 package parquet
 
-import "strings"
+import (
+	"strings"
+)
 
 type columnPath []string
 
--- a/schema.go
+++ b/schema.go
@@ -82,6 +82,13 @@ type Schema struct {
 // and the data will not be written into the parquet file(s).
 // Note that a field with name "-" can still be generated using the tag "-,".
 //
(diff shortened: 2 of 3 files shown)


achille-roussel:
We discussed offline the option to use a different struct tag name for the key, e.g. `key`, `mapkey` or `parquet-key`, which has advantages over the `key=` prefix:

- the grouping of attributes for the map key help with readability
- there is no need for repetition like there is with the `key=` prefix

Did you get a chance to give this more thoughts?

Pryz:
Did some experimentation and went with the `parquet-key` tag.

Example:
```
type Spans struct {
	Attributes map[string]json.RawMessage `parquet:"attributes,optional,json,zstd" parquet-key:"zstd"`
}
```

achille-roussel (approved):
The change is looking good!

Pryz:
Thanks for the review !

PR #270, opened 2022-07-13
buffer_test.go: Add test case for reusing writer with bloomfilters

brancz:
I don't know why yet, but this test appears to fail on arm64. Opening mainly to see if it works on amd64 and to see if anyone has an idea what the culprit might be.

@achille-roussel

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -893,9 +893,10 @@ func (c *writerColumn) flushFilterPages() error {
 		// If there is a dictionary, it contains all the values that we need to
 		// write to the filter.
 		if dict := c.dictionary; dict != nil {
-			if c.filter.bits == nil {
-				c.resizeBloomFilter(int64(dict.Len()))
-			}
+			// Need to always attempt to resize the filter, as the writer might
+			// be reused after resetting which would have reset the length of
+			// the filter to 0.
+			c.resizeBloomFilter(int64(dict.Len()))
 			if err := c.writePageToFilter(dict.Page()); err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


brancz:
If I'm getting this right I believe the problem we were seeing was that we reset the bits of bloomfilters of a column here:

https://github.com/segmentio/parquet-go/blob/6dc5e4bb634a24de7c2fa27f7c16bf1487416bdf/writer.go#L850

And because we reset the writer, the `w.filter.bits` is not nil here, and therefore it is still length 0 (because that's what it was reset to):

https://github.com/segmentio/parquet-go/blob/6dc5e4bb634a24de7c2fa27f7c16bf1487416bdf/writer.go#L896-L898

And when it is then passed to `MakeSplitBlockFilter`:

https://github.com/segmentio/parquet-go/blob/6dc5e4bb634a24de7c2fa27f7c16bf1487416bdf/bloom/filter.go#L33-L37

We create an empty filter (I checked through some logging and indeed we always create an empty filter with a reset column). So when we then attempt to insert into that empty filter we panic:

https://github.com/segmentio/parquet-go/blob/6dc5e4bb634a24de7c2fa27f7c16bf1487416bdf/bloom/filter_default.go#L12

I suspect that even on amd64 we had unsafe behavior but since the assembly code does not perform any bounds checks it just ran whatever we gave it, so it might potentially have produced garbage/caused buffer overflows before this.

achille-roussel:
Thanks for the thorough investigation! 

I'll prioritize getting a fix out today.

achille-roussel (approved):
The fix looks good, thanks for submitting it!

Thinking about the consequences here, it would have resulted in generating an empty filter. I don't think there were buffer overflow issues tho, with a buffer of length zero all hashing to find the bloom filter block would have selected "block zero", so all bits were written to the first 32 bytes of the filter slice which was retained when reusing the writer.

It was doing the wrong thing and generating to filter, but all out of range writes were still hitting the buffer for the filter, so no risk of having altered the program memory or other parts of the parquet file.

brancz:
Makes sense!

PR #284, opened 2022-07-28
add new test files

### LINKED ISSUE #274: Add latest test file from parquet-testing repository
https://github.com/apache/parquet-testing/pull/25

achille-roussel:
Fixes #274 

I added the 3 new files that were introduced in https://github.com/apache/parquet-testing recently, which triggered a few bugs in corner cases:

* The checksum computation in one of the files seems to differ with the one computed by parquet-go. I have a hard time tracking down where the issue comes from because the parquet implementations in Apache Arrow seem to not check the page CRC, and parquet-tools appears to only verify the checksum on page v1 (v2 is the latest and standard format).

* In the files that tested decoding the page index, the column index was missing for one of the columns. I don't know why this is an acceptable condition, but I updated parquet-go to handle this condition nevertheless.


file.go
```diff
@@ -619,7 +627,15 @@ func (f *filePages) readPage(header *format.PageHeader, page *dataPage, reader *
...
-		if headerChecksum != bufferChecksum {
+		// TODO: checksum validation is disabled until we figure out how the
+		// checksum of TestOpenFile/testdata/delta_length_byte_array.parquet was
+		// computed.
+		//
+		// Note that we still compute the page checksum even if we are not using
+		// to avoid skewing benchmarks.
+		//
+		// https://github.com/apache/parquet-testing/pull/24#issuecomment-1196045050
+		if false && headerChecksum != bufferChecksum {
```
kevinburkesegment, line 638: Let's pull this into a variable

```
// TODO: checksum validaiton is disabled...
const canValidateChecksums = false 
```

kevinburkesegment (approved):

PR #287, opened 2022-07-29
Add ability to set a MAP as optional via tags

### LINKED ISSUE #280: How to make an optional map?
Take the sample below:
```go
type Spans struct {
  Attributes map[string]string
}

s := parquet.SchemaOf(new(RowType))
fmt.Println(s)
```

It prints:
```
message Spans {
  required group Attributes (MAP) {
    repeated group key_value {
      required binary key (STRING);
      required binary value (STRING);
    }
  }
}
```

However, I want the Attributes map to be optional:
```
message Spans {
  optional group Attributes (MAP) {
    repeated group key_value {
      required binary key (STRING);
      required binary value (STRING);
    }
  }
}
```

I try to add a tag, but it makes the value optional, not the map.
```
type Spans struct {
	Attributes map[string]string  `parquet:"Attributes,optional"`
}
```

Could you please teach me how to do it?

Pryz:
Fixes #280 

This change update the schema API with the following:
- The tag `parquet-value` is introduce and allows the configuration of the MAP values.
- The tag `parquet` will now configure the map (top level) itself.

### DIFF
--- a/node.go
+++ b/node.go
@@ -235,13 +235,56 @@ func fieldRepetitionTypeOf(node Node) format.FieldRepetitionType {
 	}
 }
 
-type Group map[string]Node
+type Group struct {
+	nodes    map[string]Node
+	optional bool
+}
+
+type GroupNodes map[string]Node
+
+func NewGroup(n map[string]Node) Group {
+	return Group{nodes: n}
+}
+
+func groupWith(t reflect.Type, tag string) Group {
+	var (
+		group    Group
+		optional bool
+	)
+
+	setOptional := func() {
+		if optional {
+			throwInvalidNode(t, "group has multiple declaration of the optional tag", "", tag)
+		}
+		optional = true
+	}
+
+	forEachTagOption([]string{tag}, func(option, args string) {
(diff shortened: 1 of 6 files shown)


kevinburkesegment (approved):
Do we need to do anything to indicate to callers now that an empty Group struct is not usable?

node.go
```diff
@@ -235,13 +235,56 @@ func fieldRepetitionTypeOf(node Node) format.FieldRepetitionType {
 	}
 }
 
-type Group map[string]Node
+type Group struct {
+	nodes    map[string]Node
+	optional bool
+}
```
achille-roussel, line 241: Instead of changing `Group`, would it work if we used the `Optional` wrapper?
```go
Optional(Group{ ... })
```

Pryz: mmm will take a look but we are gonna have a type issues with `mapNode` which is expecting a Group and not a Node right ?

Pryz: Actually we can vary the tag there. Thanks ! Will refactor :)

PR #291, opened 2022-08-01
optimize DELTA_LENGTH_BYTE_ARRAY decoding

achille-roussel:
Follow up to #289, this PR optimizes the decoding of DELTA_LENGTH_BYTE_ARRAY pages.

We get ~2x better throughput with this change compared to the parent branch:
```
name                                       old time/op    new time/op    delta
Decode/DELTA_LENGTH_BYTE_ARRAY/byte_array    24.0µs ± 0%    12.0µs ± 0%   -50.07%  (p=0.000 n=10+10)

name                                       old speed      new speed      delta
Decode/DELTA_LENGTH_BYTE_ARRAY/byte_array  4.53GB/s ± 0%  9.07GB/s ± 0%  +100.28%  (p=0.000 n=10+10)

name                                       old value/s    new value/s    delta
Decode/DELTA_LENGTH_BYTE_ARRAY/byte_array      416M ± 0%      834M ± 0%  +100.29%  (p=0.000 n=10+10)
```

### DIFF
--- a/encoding/delta/length_byte_array_amd64.s
+++ b/encoding/delta/length_byte_array_amd64.s
@@ -12,13 +12,14 @@ TEXT ·decodeByteArrayLengths(SB), NOSPLIT, $0-56
     XORQ DI, DI // invalidLength
     XORQ SI, SI
 
-    CMPQ BX, $4
-    JB test
+    CMPQ CX, $4
+    JL test
 
-    MOVQ CX, R9
-    SHRQ $2, R9
-    SHLQ $2, R9
+    MOVQ CX, R8
+    SHRQ $2, R8
+    SHLQ $2, R8
 
+    MOVL $0, (AX)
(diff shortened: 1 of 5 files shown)


encoding/delta/length_byte_array_amd64.s
```diff
@@ -0,0 +1,88 @@
...
+    JL test
+
+    MOVQ CX, R8
+    SHRQ $2, R8
+    SHLQ $2, R8
+
+    MOVL $0, (AX)
+    PXOR X0, X0
+    PXOR X3, X3
+    // This loops computes the prefix sum of the lengths array in order to
```
kevinburkesegment, line 25:
```suggestion
    // This loop computes the prefix sum of the lengths array in order to
```

kevinburkesegment (approved):

PR #293, opened 2022-08-01
Add the parquet-value tag and add the ability to make a map optional

### LINKED ISSUE #280: How to make an optional map?

Pryz:
Fixes #280

### DIFF
--- a/schema.go
+++ b/schema.go
@@ -84,9 +84,9 @@ type Schema struct {
 //
 // The configuration of Parquet maps are done via two tags:
 //   - The `parquet-key` tag allows to configure the key of a map.
-//   - The `parquet-value` tag allows to configure the values of a map.
+//   - The parquet-value tag allows users to configure a map's values, for example to declare their native Parquet types.
 //
-// When configuring a Parquet map, the `parquet` tag will configure map itself.
+// When configuring a Parquet map, the `parquet` tag will configure the map itself.
 //
 // For example, the following will set the int64 key of the map to be a timestamp:
 //
--- a/schema.go
+++ b/schema.go
@@ -539,8 +539,18 @@ func nodeOf(t reflect.Type, tag []string) Node {
 		n = Map(
 			makeNodeOf(t.Key(), t.Name(), []string{keyTag}),
 			makeNodeOf(t.Elem(), t.Name(), []string{valueTag}),
-			mapTag,
 		)
+
+		forEachTagOption([]string{mapTag}, func(option, args string) {
+			switch option {
+			case "":
+				return
+			case "optional":
+				n = Optional(n)
+			default:
+				throwUnknownTag(t, "map", option)
+			}
+		})
 
 	case reflect.Struct:
 		return structNodeOf(t)
--- a/schema.go
+++ b/schema.go
@@ -82,8 +82,11 @@ type Schema struct {
 // and the data will not be written into the parquet file(s).
 // Note that a field with name "-" can still be generated using the tag "-,".
 //
-// Map key configuration can be pass using the `parquet-key` tag. This tag will follow the
-// same properties than the `parquet` tag.
+// The configuration of Parquet maps are done via two tags:
+//   - The `parquet-key` tag allows to configure the key of a map.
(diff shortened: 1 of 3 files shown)


schema.go
```diff
@@ -82,8 +82,11 @@ type Schema struct {
 // and the data will not be written into the parquet file(s).
 // Note that a field with name "-" can still be generated using the tag "-,".
 //
-// Map key configuration can be pass using the `parquet-key` tag. This tag will follow the
-// same properties than the `parquet` tag.
+// The configuration of Parquet maps are done via two tags:
+//   - The `parquet-key` tag allows to configure the key of a map.
+//   - The `parquet-value` tag allows to configure the values of a map.
```
kevinburkesegment, line 87: Not quite a sentence... maybe "The parquet-value tag allows users to configure a map's values, for example to declare their native Parquet types."

schema.go
```diff
@@ -82,8 +82,11 @@ type Schema struct {
...
+//
+// When configuring a Parquet map, the `parquet` tag will configure map itself.
```
kevinburkesegment, line 89:
```suggestion
// When configuring a Parquet map, the `parquet` tag will configure the map itself.
```

kevinburkesegment: The tests reference e.g. `timestamp(microsecond)` - maybe we can provide a list of valid values that users can provide, or more examples?

schema_test.go
```diff
@@ -137,6 +137,48 @@ func TestSchemaOf(t *testing.T) {
...
+
+		{
+			value: new(struct {
+				A map[int64]string `parquet:",optional"`
+			}),
+			print: `message {
+	optional group A (MAP) {
+		repeated group key_value {
+			required int64 key (INT(64,true));
+			required binary value (STRING);
```
kevinburkesegment, line 165: I guess I'd expect the ",optional" bit to make this field not "required"? Or does optional imply something else?

kevinburkesegment (approved):
I hope we can extend the docs a little bit to explain what tag values are available and what they do, but the code looks OK.

schema_test.go
Pryz: if you want the value field to be optional you will need :
```
			value: new(struct {
				A map[int64]string `parquet:",optional" parquet-value:",optional"`
			}),
```

type.go
```diff
@@ -1453,13 +1453,26 @@ func (t *listType) Decode(dst, _ []byte, _ encoding.Encoding) ([]byte, error) {
 // Map constructs a node of MAP logical type.
 //
 // https://github.com/apache/parquet-format/blob/master/LogicalTypes.md#maps
-func Map(key, value Node) Node {
-	return mapNode{Group{
+func Map(key, value Node, tag ...string) Node {
```
achille-roussel, line 1456: This model looks a bit odd since it's mixing the struct tags and the manual construction of nodes.

How about keeping this logic in `nodeOf` ?

schema.go
Pryz: You mean for the precision of the timestmap, It is there already https://github.com/segmentio/parquet-go/blob/ffd0eebbd2b0668e5e2ce0834a649dca7ae74fcc/schema.go#L65

type.go
Pryz: yeah I know something was weird. updated. thanks

achille-roussel (approved):

PR #295, opened 2022-08-02
added string represented date/timestamp

vc42-2:


utils/utils.go
```diff
@@ -0,0 +1,62 @@
+package utils
```
Pryz, line 1: Since these transformations are only used in the parquet package, we should probably move them there and not expose them. Each public function we expose as a maintenance cost which is unnecessary in this case.

utils/utils.go
```diff
@@ -0,0 +1,62 @@
...
+
+import (
+	"fmt"
+	"strings"
+	"time"
+)
+
+func StringToTime(ts string) *time.Time {
+	layout := "2006-01-02 15:04:05.000-07"
```
Pryz, line 10: This looks a bit restrictive to only allow a specific format/layout. Could we get more flexibility by letting the user defining the format ?

Like, for example:
```
type Message struct {
  T string `parquet:",timestamp(microsecond,layout=rfc3339)"
}
```

Pryz:
Any thoughts about my comments ?

utils/utils.go
vc42-2: I agree about format, just tied up at work at the moment..

achille-roussel:
Hello @vc42-2, do you think a solution like #321 would satisfy your use case?

vc42-2:
Yess, it will, thanks.

Issue #248, opened 2022-06-22
Optimise Parquet ReadRow


annanay25:
At the moment, we perform one read per column to seek to a given row within a page. Below is a screenshot of the readrow operations for a schema with ~40 fields and ReadRow latency linearly increases with the number of fields. 

![image](https://user-images.githubusercontent.com/10982987/175005687-f769bfaf-a2ca-4b0a-b5cb-2c94a576d3da.png)

One way to improve this is to move markers for multiple columns concurrently -- lets use this issue to gather other ideas if any!

zolstein:
@annanay25 Could you share any more information about how you generated that graph? Ideally a script / test-file, but failing that just a description of what you did and maybe some pseudocode? (In particular I'm curious how you set up the file to read from S3.) I have some thoughts on how the read performance might be improved based on some ideas / work from the [xitongsys library](https://github.com/xitongsys/parquet-go), but before I dump my brain and possibly hijack the issue I'd like to know what specific use-case you want to optimize.

annanay25:
Hi @zolstein, thanks for looking into this. The visualisation is that of a trace exported from an instrumented application (https://github.com/grafana/tempo). `parquet.backendBlock.FindTraceByID` (the top line in the trace view) is a span generated from the [FindTraceByID operation](https://github.com/grafana/tempo/blob/150eeea11d11d8231d172c9fcebb8fbfbaaab35b/tempodb/encoding/vparquet/block_findtracebyid.go#L160) performed on our Query service responsible for fetching the parquet row corresponding to a given ID. The service scans through records in a parquet file present on object storage - first it sifts through the ID column to match the requested ID, and once found it retrieves the corresponding row to be displayed to the user. The resulting waterfall diagram in the above trace view shows the sequential nature of reads performed for a `ReadRow` operation. Since we have about 40-50 fields in the parquet schema present on the file, the latency for a single readrow operation goes up to 2.58s as shown.

> In particular I'm curious how you set up the file to read from S3

As `parquet.OpenFile()` takes an `io.ReaderAt` as an input, we implemented a wrapper around the s3/gcs/azure file handler that implements the `io.ReaderAt` interface. [Here](https://github.com/grafana/tempo/blob/150eeea11d11d8231d172c9fcebb8fbfbaaab35b/tempodb/encoding/vparquet/readers.go#L14-L34) is the code for reference.

zolstein:
[...]

PR #297, opened 2022-08-03
Refactor reading of parquet pages

achille-roussel:
I'm going to use this as the base branch to work on https://github.com/segmentio/parquet-go/issues/248. As a first step, I am removing optimizations that make the internal logic more complex. By simplifying the implementation, introducing changes to address the issue will be simpler.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -1013,13 +1013,13 @@ func (c *writerColumn) WritePage(page Page) (numValues int64, err error) {
 				})
 				return numValues, err
 
-			case CompressedPage:
+				//case CompressedPage:
 				// Compressed pages are written as-is to the compressed page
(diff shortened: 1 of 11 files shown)


writer.go
```diff
@@ -1013,13 +1013,13 @@ func (c *writerColumn) WritePage(page Page) (numValues int64, err error) {
 				})
 				return numValues, err
 
-			case CompressedPage:
+				//case CompressedPage:
```
kevinburkesegment, line 1016: ?

kevinburkesegment (approved):

writer.go
achille-roussel: See the comment below explaining why this code is disabled for now.

PR #298, opened 2022-08-04
Upgrade to Go 1.19

Pryz:


file.go
```diff
@@ -147,19 +147,19 @@ func OpenFile(r io.ReaderAt, size int64, options ...FileOption) (*File, error) {
 // Only leaf columns have indexes, the returned indexes are arranged using the
 // following layout:
 //
-//	+ -------------- +
```
kevinburkesegment, line 150: I don't think this is correct... we ran into this problem else where too. I think we might need to avoid plus signs in the table.

kevinburkesegment (approved):

file.go
kevinburkesegment: It's just supposed to look like a table in the text view - godoc HTML has no support for tables.

PR #300, opened 2022-08-05
Parallel I/O

achille-roussel:
This PR adds a new `pio` package, along with APIs to support performing parallel I/O on parquet files. The intent is to be able to amortize the cost of I/O latency when reading multiple file sections (e.g. when loading pages from multiple columns).

I am opening this PR against #297, it only is a building block that I intend to use to address performance issues when reading parquet files.

The key change is the introduction of `pio.MultiReadAt(io.ReaderAt, []Op)`, which is the core API that the implementation will rely on. I also added platform specific implementations of this API, leveraging async I/O operations on Linux and Darwin, as well as a generic fallback mechanism using the Go runtime. Finally, I added an extension mechanism supported by implementation of the `pio.ReaderAt` interface, and a test suite in the `pio/piotest` package to validate the behavior of custom implementations of that interface.


pio/file_linux.go
```diff
@@ -0,0 +1,283 @@
...
+
+type ioctx struct {
+	ctx io_context_t
+	ptr []*iocb
+	req []iocb
+	res []io_event
+}
+
+const (
+	// Hard limits, chosen arbitrarily, we should revisit if they are not
```
Pryz, line 156: Should we make those configuration so that we can experiment with those values without recompiling things ?

Pryz (approved):
Very elegant ! Wondering if we should expose some stats (mostly queue size) to monitor contentions. 

Looks very similar to something like io_uring but with a LIFO instead :)

pio/file_linux.go
```diff
@@ -0,0 +1,283 @@
...
+func io_setup(n int) (io_context_t, syscall.Errno) {
+	ctx := io_context_t(0)
+	_, _, errno := syscall.Syscall(syscall.SYS_IO_SETUP, uintptr(n), uintptr(unsafe.Pointer(&ctx)), 0)
+	return ctx, errno
+}
+
+func io_submit(ctx io_context_t, reqs []*iocb) (int, syscall.Errno) {
+	p := unsafe.Pointer(&reqs[0])
+	n := len(reqs)
+	r, _, errno := syscall.Syscall(syscall.SYS_IO_SUBMIT, uintptr(ctx), uintptr(n), uintptr(p))
```
kevinburkesegment, line 261: Should you use `x/sys/unix` here? I believe the syscall package is frozen.

```
r, _, errno := unix.Syscall(unix.SYS_IO_SUBMIT, uintptr(ctx), uintptr(n), uintptr(p))
```

pio/file_linux.go
```diff
@@ -0,0 +1,283 @@
...
+	return int(r), errno
+}
+
+func io_getevents(ctx io_context_t, min int, events []io_event) (int, syscall.Errno) {
+	p := unsafe.Pointer(&events[0])
+	n := len(events)
+	r, _, errno := syscall.Syscall6(syscall.SYS_IO_GETEVENTS, uintptr(ctx), uintptr(min), uintptr(n), uintptr(p), 0, 0)
```
kevinburkesegment, line 268: Same comment here about x/sys/unix

kevinburkesegment (approved):

achille-roussel:
I'm going to close this, abandoning this approach in favor of #301

PR #301, opened 2022-08-08
Concurrent column page reader

achille-roussel:
This PR is an alternative to #300, which should be a more general way to address #248 

The implementation uses a separate goroutine to read column pages, allowing the processing to happen concurrently.

A potential downside is that we might read a bit more data than necessary, as we would likely consume two pages per column minimum, even tho only a single page might be needed (e.g. if just one row is being read). We can figure out how to optimize this later I think.

Goroutines introduce concurrency and complexity, but they also integrate well with the Go language, which I believe is the trade off we're working with here. Code extensibility and simplicity is important, using parallel I/O will be more complex and make it more difficult to integrate with existing Go code.

### DIFF
--- a/page.go
+++ b/page.go
@@ -2,9 +2,11 @@ package parquet
 
 import (
 	"bytes"
+	"context"
 	"fmt"
 	"io"
 	"runtime"
+	"runtime/pprof"
 	"sync"
 
 	"github.com/segmentio/parquet-go/deprecated"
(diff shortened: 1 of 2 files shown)


Pryz (approved):

achille-roussel:
I revisited the implementation in the last commits to improve concurrency and avoid a slow-start behavior caused by lazily starting the underlying goroutine in ReadPage.

page.go
```diff
@@ -126,23 +129,144 @@ type Pages interface {
 	io.Closer
 }
 
+// AsyncPages wraps the given Pages instance to perform page reads
+// asynchronoulsy in a separate goroutine.
```
kevinburkesegment, line 133:
```suggestion
// asynchronously in a separate goroutine.
```

page.go
```diff
@@ -126,23 +129,144 @@ type Pages interface {
...
+//
+// Performing page reads asynchronously is important when the application may
+// be reading pages from a high latency backend, and processing of the last
```
kevinburkesegment, line 136:
```suggestion
// be reading pages from a high latency backend, and the last
```

PR #315, opened 2022-08-22
Add guard code in sparse array `offset()`

### LINKED ISSUE #299: Occasional crash during garbage collection
Very rare crashes during GC, hard to reproduce.  Looks like the culprit is sparse/array.go [func (a array) offset(off uintptr) array].  Possibly when "a.ptr == nil" but "off > 0", the GC sees a broken ptr. 

 One possible fix (fix 1):
```diff
--- a/sparse/array.go
+++ b/sparse/array.go
@@ -55,6 +55,18 @@ func (a array) slice(i, j int) array {
 }
 
 func (a array) offset(off uintptr) array {
+       if a.off == 0 {
+               off = 0
+               a.len = 0
+       } else {
+               offsetSize := off / a.off
+               if a.len > offsetSize {
+                       a.len -= offsetSize
+               } else {
+                       off = 0
+                       a.len = 0
+               }
+       }
        return array{
                ptr: unsafe.Add(a.ptr, off),
                len: a.len,
```
Another possible fix (fix 2):
```diff
--- a/sparse/array.go
+++ b/sparse/array.go
@@ -55,6 +55,9 @@ func (a array) slice(i, j int) array {
 }
 
 func (a array) offset(off uintptr) array {
+       if a.ptr == nil {
+               return a
+       }
        return array{
                ptr: unsafe.Add(a.ptr, off),
                len: a.len,
```
Not sure if the "off"argument can apply to the whole array or only the first row.  Depending on that, one can use fix 1 or fix 2.  We use fix 1, and GC related crashes have not been happening  so far.

joe-elliott:
This code was suggested by @vc42-2. We have tested it in an environment writing 66k rows/second. Without this change we would see this panic ~10-20 times an hour. This change has been running for ~2 hours with no GC crashes.

I have no idea if this fix is appropriate or not :), but I'd like to move the conversation forward on this issue. It is currently blocking the very nice performance improvements gained by moving to `GenericWriter`.

Fixes #299.

@vc42-2 Heads up that I submitted this fix. This is your code and I'd be glad to drop this PR in favor of your own.

### DIFF
--- a/sparse/array.go
+++ b/sparse/array.go
@@ -55,6 +55,19 @@ func (a array) slice(i, j int) array {
 }
 
 func (a array) offset(off uintptr) array {
+	if a.off == 0 {
+		off = 0
+		a.len = 0
+	} else {
+		offsetSize := off / a.off
+		if a.len > offsetSize {
+			a.len -= offsetSize
+		} else {
+			off = 0
+			a.len = 0
+		}
+	}
+
 	return array{
 		ptr: unsafe.Add(a.ptr, off),
 		len: a.len,


achille-roussel:
We merged https://github.com/segmentio/parquet-go/pull/323 as the fix which should have addressed the original issue.

Thanks a lot for reporting and your contributions in the investigation, please let me know if you are still experiencing this problem with the latest version of parquet-go.

PR #316, opened 2022-08-23
Write page statistics to file

zsolt-haraszti:
* Copy page stats from page headers to chunk metadata so it ends up in
  the file trailer.

* Test case to verify presence of statistics in a generated file

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -1265,11 +1265,14 @@ func (c *writerColumn) recordPageStats(headerSize int32, header *format.PageHead
 
 	pageType := header.Type
 	encoding := format.Encoding(-1)
+	var statistics format.Statistics
 	switch pageType {
 	case format.DataPageV2:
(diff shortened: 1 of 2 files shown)


achille-roussel:
Hello @zsolt-haraszti, thanks for submitting a pull request.

The statistics are not generated unless the application opts-in to do so https://github.com/segmentio/parquet-go/blob/3f82efce98078b7448f5cf0e41ae9aa8927bfd31/writer.go#L1098

We made it optional because the latest parquet format recommends not to write page statistics as part of the page header since the page index has the same information in a more usable form. This means that in the current form, the option would also control the creation of column chunk statistics.

On a different note, I believe the change you submitted may have an issue with the correctness of statistics set on the column chunk metadata. There may be multiple pages per column chunk, so the column chunk statistics should reflect the aggregate over all pages rather than hold the statistics of the last page written to the chunk.

Let me know if you have any comments on the feedback.

PR #319, opened 2022-08-26
Refactor Conversion Code to be schema-aware

### LINKED ISSUE #250: Schema conversion fails with error "missing optional column <x>"
After adding a new uint64 column to our [go struct](https://github.com/grafana/tempo/blob/parquet/tempodb/encoding/vparquet/schema.go#L138), we are unable to read data files written previously without this column, using `parquet.Reader.Read(struct)`.  The library detects the schema drift, and attempts to automatically convert, but fails with an unexpected error `rs → Resource → Attrs → Value → row is missing optional column 2`.  

Was able to create a minimal reproduction in the existing conversion tests.  It seems related to a schema with at least 2 levels of nesting and repeated columns.  In this test, we populate an AddressBook, then try to convert it with a new field `Extra string`. 

```go
var conversionTests = [...]struct {
	scenario string
	from     interface{}
	to       interface{}
}{
....
	{
		scenario: "extra column on complex struct",

		from: AddressBook{
			Owner: "Julien Le Dem",
			Contacts: []Contact{
				{
					Name:        "Dmitriy Ryaboy",
					PhoneNumber: "555 987 6543",
				},
				{
					Name: "Chris Aniszczyk",
				},
			},
		},

		to: struct {
			AddressBook
			Extra string
		}{
			AddressBook: AddressBook{
				Owner: "Julien Le Dem",
				Contacts: []Contact{
					{
						Name:        "Dmitriy Ryaboy",
						PhoneNumber: "555 987 6543",
					},
					{
						Name: "Chris Aniszczyk",
					},
				},
			},
		},
	},
}
```

```
$ go test -v -run=TestConvert
=== RUN   TestConvert
=== RUN   TestConvert/convert_between_rows_which_have_the_same_schema
=== RUN   TestConvert/missing_column
=== RUN   TestConvert/missing_optional_column
=== RUN   TestConvert/missing_repeated_column
=== RUN   TestConvert/extra_column
=== RUN   TestConvert/extra_optional_column
=== RUN   TestConvert/extra_repeated_column
=== RUN   TestConvert/extra_column_on_complex_struct
    convert_test.go:156: contacts → phoneNumber → row is missing optional column 3
--- FAIL: TestConvert (0.00s)
    --- PASS: TestConvert/convert_between_rows_which_have_the_same_schema (0.00s)
    --- PASS: TestConvert/missing_column (0.00s)
    --- PASS: TestConvert/missing_optional_column (0.00s)
    --- PASS: TestConvert/missing_repeated_column (0.00s)
    --- PASS: TestConvert/extra_column (0.00s)
    --- PASS: TestConvert/extra_optional_column (0.00s)
    --- PASS: TestConvert/extra_repeated_column (0.00s)
    --- FAIL: TestConvert/extra_column_on_complex_struct (0.00s)
FAIL
```

### LINKED ISSUE #308: Error converting rows with repeated groups
When parsing a file with repeated groups, if the read schema does not _exactly_ match the file schema, the reader fails to parse the file. This seems to happen because when `(*convertRows).ReadRows` converts values between the source and target column, it implicitly sorts the row by column index, but `(*Schema).Reconstruct` expects the columns of repeated nested groups to have their fields interleaved by sub-object.

This problem occurs whether adding or removing a field in the read schema.

Sample test-case:
```go
func TestConvertNested(t *testing.T) {
	type Nested struct {
		N1 int64
		N2 int64
	}
	type WriteStruct struct {
		F1 int64
		F2 []Nested
	}
	type ReadStruct struct {
		F2 []Nested
	}
	var buf bytes.Buffer
	err := parquet.Write[WriteStruct](&buf, []WriteStruct{
		{F1: 1, F2: []Nested{{1, 2}, {3, 4}}},
	})
	if err != nil {
		t.Fatal(err)
	}
	rows, err := parquet.Read[ReadStruct](bytes.NewReader(buf.Bytes()), int64(buf.Len()))
	if err != nil {
		// Fails here with: F2 → N2 → no values found in parquet row for column 1
		t.Fatal(err)
	}
	require.Equal(t, []ReadStruct{{[]Nested{{1, 2}, {3, 4}}}}, rows)
}
```

abraithwaite:
Refactor the conversion code to be schema-aware and prepare the rows in a manner that the reconstruct functions expect.

Fixes #308
Fixes #250


convert.go
```diff
@@ -82,6 +84,150 @@ func (c *conversion) putBuffer(b *conversionBuffer) {
...
+
+// The following may be unecessary
+// func (c *conversion) convertOptional(node Node, target, source Row) (Row, error) {
+// }
+
+// func (c *conversion) convertMap(node Node, target, source Row) (Row, error) {
+// }
+
+// func (c *conversion) convertList(node Node, target, source Row) (Row, error) {
+// }
```
abraithwaite: I think these maybe unnecessary, but need to confirm.

convert.go
```diff
@@ -82,6 +84,150 @@ func (c *conversion) putBuffer(b *conversionBuffer) {
...
+
+func (c *conversion) Convert2(target, source Row) (Row, error) {
```
abraithwaite: Replace the original convert function.

convert.go
```diff
@@ -178,12 +325,14 @@ func Convert(to, from Node) (conv Conversion, err error) {
...
 
-	return &conversion{
+	c := &conversion{
 		targetColumnKinds:   targetColumnKinds,
 		targetToSourceIndex: targetToSourceIndex,
 		sourceToTargetIndex: sourceToTargetIndex,
 		schema:              schema,
-	}, nil
+	}
+	c.convertFunc = makeConvertFunc(c, to)
```
abraithwaite, line 280: This is a bit hacky. How do you use `conversion` anywhere in the convert functions?

convert_test.go
```diff
@@ -97,10 +103,162 @@ var conversionTests = [...]struct {
...
+
+// 			value = value.Elem()
+// 			if !reflect.DeepEqual(value.Interface(), test.to) {
+// 				t.Errorf("converted value mismatch:\nwant = %+v\ngot  = %+v", test.to, value.Interface())
+// 			}
+// 		})
+// 	}
+// }
+
+func TestConvert2(t *testing.T) {
```
abraithwaite: This additional test is identical to the other.  Cleanup

notes.txt
```diff
@@ -0,0 +1,44 @@
+
```
abraithwaite, line 1: Remove notes.

row.go
```diff
@@ -521,14 +530,26 @@ func reconstructFuncOf(columnIndex int16, node Node) (int16, reconstructFunc) {
 
 //go:noinline
 func reconstructFuncOfOptional(columnIndex int16, node Node) (int16, reconstructFunc) {
+	// What's the purpose of forcing the passed in node to Required for the Optional fields?
+	// Ah, I see it's so that we eventually reach the leaf base-case?
```
abraithwaite, line 534: Let's confirm and add a real comment here, since this is confusing.

row.go
```diff
@@ -521,14 +530,26 @@ func reconstructFuncOf(columnIndex int16, node Node) (int16, reconstructFunc) {
...
 	nextColumnIndex, reconstruct := reconstructFuncOf(columnIndex, Required(node))
+	// What if this is the last node/column in the row? We just incremented the
+	// nextColumnIndex meaning that if this is the last one in the list,
+	// rowLength will always be > 0 here, triggering the second if statement in
+	// the closure below, even if the optional field is empty
 	rowLength := nextColumnIndex - columnIndex
```
abraithwaite, line 540: I still don't understand why this is here and don't have it in the convert code yet.

row.go
```diff
@@ -521,14 +530,26 @@ func reconstructFuncOf(columnIndex int16, node Node) (int16, reconstructFunc) {
...
+
 	return nextColumnIndex, func(value reflect.Value, levels levels, row Row) (Row, error) {
-		if !row.startsWith(columnIndex) {
+		if len(row) > 0 && !row.startsWith(columnIndex) {
```
abraithwaite, line 543: Remove if I don't know why I added it.

row.go
```diff
@@ -599,6 +625,7 @@ func reconstructRepeated(columnIndex, rowLength int16, levels levels, row Row, d
 
 	var err error
 	for row.startsWith(columnIndex) && row[0].repetitionLevel == levels.repetitionLevel {
+		// TODO: check specific error here.
```
abraithwaite, line 628: We should hopefully be getting a clear error that we're looking for to break the loop.  Feels dangerous to catch them all here otherwise.

row.go
achille-roussel: The intent is to avoid infinite recursion, otherwise the `reconstructFuncOf` call below would call `reconstructFuncOfOptional` again.

schema.go
```diff
@@ -158,6 +158,7 @@ func makeReconstructFunc(node Node) (reconstruct reconstructFunc) {
 	if !node.Leaf() {
 		_, reconstruct = reconstructFuncOf(0, node)
 	}
+	// will this ever return nil?  Implicit assignment as named return parameter
```
achille-roussel, line 161: It would return `nil` if called with a non-leaf node I believe.

abraithwaite:
Thanks for the clarifications!

convert.go
```diff
@@ -82,10 +83,103 @@ func (c *conversion) putBuffer(b *conversionBuffer) {
...
+	nextIdx, convFunc := c.convertFuncOf(tgtIdx, Required(node))
+	return nextIdx, func(tgt Row, src *conversionBuffer) (Row, error) {
+		var err error
+
+		for i, elem := range src.columns[tgtIdx] {
+			// if our repetitionLevel is not equal to the definition level on our
+			// second or subsequent copies, then this element belongs to another
+			// repeated object above us.
+			if i != 0 && int16(elem.repetitionLevel) != int16(elem.definitionLevel) {
+				break
```
abraithwaite, line 163: I'm pretty sure this logic holds 100% of the time, but would appreciate a second opinion here!

achille-roussel: Maybe a counter example would be a case of this column structure:
```
message M {
  repeated group A {
    optional int64 B;
  }
}
```
In which case values of the column A.B may have a definition level that differs from the repetition level?
- a null value in A.B would have definition level 1
- a non-null value in A.B would have definition level 2
- the repetition level is 0 on the value of the first record, and 2 on the other values

schema.go
```diff
@@ -259,10 +260,16 @@ func (s *Schema) Reconstruct(value interface{}, row Row) error {
 		v = v.Elem()
 	}
 	var err error
+	// TODO: when is s.reconstruct nil? Should that error?
```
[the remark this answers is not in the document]
abraithwaite: Added a panic in an else statement

convert.go
abraithwaite: I'll see if I can't produce a failing test case for this scenario, or otherwise come up with a good explanation of the behavior in this case.

abraithwaite: Got a failing test case, we'll see what I can do with this.  Might just have to pass in the levels object, we'll see!

achille-roussel: Keeping track of the expected repetition level is probably gonna be an effective strategy (this is what we do in general across the library).

abraithwaite: Seems like there's no way around it from what I can tell

achille-roussel (approved):
Assuming you want to keep all the comments you added (some are questions it seems), this looks good to go 👍

abraithwaite:
> Assuming you want to keep all the comments you added (some are questions it seems), this looks good to go 👍

Good catch, I meant to update those comments with the answers.  Will do that then merge! :-)

PR #320, opened 2022-08-26
Configurable read buffer

joe-elliott:
The default read buffer size is standard for disk access, but we are using this library with object storage backing the parquet.File. With object storage hundreds of tiny reads of 4096 is obviously a non-starter so we built a [simple buffer](https://github.com/grafana/tempo/blob/main/pkg/io/buffered.go). Unfortunately this buffer often overreads by quite a bit and for performance sensitive accesses we have found that simply increasing `defaultBufferSize` in this library beats our buffer in nearly every case.

This PR makes the `defaultBufferSize` configurable to allow those who are backing their `parquet.File`s with unconventional storage some control over how it is accessed. It also removes `defaultLevelBufferSize` because it was unused. No issue with re-adding if it's there for a reason I don't understand.

Finally, I was unable to write a convincing test for this configuration option. Any advice on how to proceed would be appreciated.

### DIFF
--- a/file.go
+++ b/file.go
@@ -30,6 +30,7 @@ type File struct {
 	columnIndexes []format.ColumnIndex
 	offsetIndexes []format.OffsetIndex
 	rowGroups     []RowGroup
+	cfg           *FileConfig
 }
 
 // OpenFile opens a parquet file and reads the content between offset 0 and the given
@@ -40,11 +41,11 @@ type File struct {
 // a file does not validate that the pages have valid checksums.
 func OpenFile(r io.ReaderAt, size int64, options ...FileOption) (*File, error) {
 	b := make([]byte, 8)
-	f := &File{reader: r, size: size}
 	c, err := NewFileConfig(options...)
 	if err != nil {
 		return nil, err
 	}
+	f := &File{reader: r, size: size, cfg: c}
 
 	if _, err := r.ReadAt(b[:4], 0); err != nil {
 		return nil, fmt.Errorf("reading magic header of parquet file: %w", err)
@@ -464,12 +465,15 @@ type filePages struct {
 	index      int
 	skip       int64
 	dictionary Dictionary
+
+	bufferSize int
 }
 
 func (f *filePages) init(c *fileColumnChunk) {
 	f.chunk = c
 	f.baseOffset = c.chunk.MetaData.DataPageOffset
 	f.dataOffset = f.baseOffset
+	f.bufferSize = c.file.cfg.ReadBufferSize
 
 	if c.chunk.MetaData.DictionaryPageOffset != 0 {
 		f.baseOffset = c.chunk.MetaData.DictionaryPageOffset
@@ -477,7 +481,7 @@ func (f *filePages) init(c *fileColumnChunk) {
 	}
 
 	f.section = *io.NewSectionReader(c.file, f.baseOffset, c.chunk.MetaData.TotalCompressedSize)
-	f.rbuf = getBufioReader(&f.section)
+	f.rbuf = getBufioReader(&f.section, f.bufferSize)
 	f.decoder.Reset(f.protocol.NewReader(f.rbuf))
 }
 
@@ -544,7 +548,7 @@ func (f *filePages) ReadPage() (Page, error) {
 
 func (f *filePages) readDictionary() error {
 	chunk := io.NewSectionReader(f.chunk.file, f.baseOffset, f.chunk.chunk.MetaData.TotalCompressedSize)
-	rbuf := getBufioReader(chunk)
+	rbuf := getBufioReader(chunk, f.bufferSize)
 	defer putBufioReader(rbuf)
 
 	decoder := thrift.NewDecoder(f.protocol.NewReader(rbuf))
@@ -700,10 +704,10 @@ var (
 	bufioReaderPool sync.Pool
 )
 
-func getBufioReader(r io.Reader) *bufio.Reader {
+func getBufioReader(r io.Reader, bufferSize int) *bufio.Reader {
 	rbuf, _ := bufioReaderPool.Get().(*bufio.Reader)
 	if rbuf == nil {
-		rbuf = bufio.NewReaderSize(r, defaultReadBufferSize)
+		rbuf = bufio.NewReaderSize(r, bufferSize)
 	} else {
 		rbuf.Reset(r)
 	}
--- a/file.go
+++ b/file.go
@@ -549,7 +549,7 @@ func (f *filePages) ReadPage() (Page, error) {
 func (f *filePages) readDictionary() error {
 	chunk := io.NewSectionReader(f.chunk.file, f.baseOffset, f.chunk.chunk.MetaData.TotalCompressedSize)
 	rbuf := getBufioReader(chunk, f.bufferSize)
-	defer putBufioReader(rbuf, f.bufferSize)
+	defer putBufioReader(rbuf)
 
 	decoder := thrift.NewDecoder(f.protocol.NewReader(rbuf))
 	header := new(format.PageHeader)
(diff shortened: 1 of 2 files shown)


file.go
```diff
@@ -31,6 +30,7 @@ type File struct {
 	columnIndexes []format.ColumnIndex
 	offsetIndexes []format.OffsetIndex
 	rowGroups     []RowGroup
+	cfg           *FileConfig
```
achille-roussel, line 33: We used the word `config` rather than `cfg` in other types, what do you think about renaming this field to keep it consistent?

file.go
```diff
@@ -701,10 +704,10 @@ var (
 	bufioReaderPool sync.Pool
 )
 
-func getBufioReader(r io.Reader) *bufio.Reader {
+func getBufioReader(r io.Reader, bufferSize int) *bufio.Reader {
```
achille-roussel, line 707: An issue with this approach is that we end up with a single pool of buffers with different sizes. When the pool fills up, the request buffer size ends up not being respected (unless the application uses a single buffer size).

An option could be to maintain multiple pools for different buffer sizes (e.g. multiple of 4096, up to a sane limit). Then we round up the requested buffer size to the nearest multiple to offer a better guarantee to the application that we will be respecting the requested size.

What do you think?

joe-elliott: It did not occur to me that this was a shared pool that all parquet.File's access. Good catch. I consider it reasonable to assume that most applications would have at most 2 or 3 different buffer sizes throughout. We could maintain a `map[buffer_size] -> sync.Pool` ?

achille-roussel: Sounds good 👍

file.go
joe-elliott: +1, done

file.go
```diff
@@ -698,22 +701,32 @@ func (f *filePages) columnPath() columnPath {
 }
 
 var (
-	bufioReaderPool sync.Pool
+	bufioReaderPool = map[int]*sync.Pool{}
```
achille-roussel, line 704: We need access to this map to be synchronized on a mutex, or use a `sync.Map` to prevent race conditions when multiple goroutines are using the map.

joe-elliott: oof, that seems obvious in retrospect. thank you.

joe-elliott:
Added a RWMutex protected map from buffer sizes -> sync.Pool. Using the following benchmark:
```
func BenchmarkBufioGetPut(b *testing.B) {
	for i := 0; i < b.N; i++ {
		rdr := getBufioReader(nil, 100)
		putBufioReader(rdr)
	}
}
```

Compared to main (sync.Map was even slower):
```
$ benchstat before.txt after.txt 
name           old time/op    new time/op    delta
BufioGetPut-8    29.2ns ± 4%    59.0ns ± 2%  +101.81%  (p=0.000 n=10+10)

name           old alloc/op   new alloc/op   delta
BufioGetPut-8     0.00B          0.00B           ~     (all equal)

name           old allocs/op  new allocs/op  delta
BufioGetPut-8      0.00           0.00           ~     (all equal)
```

I would like to write tests against these methods but all of the testing is currently in external packages. Should I export these methods for testing? or just skip it?

file.go
```diff
@@ -714,6 +718,22 @@ func getBufioReader(r io.Reader) *bufio.Reader {
...
+	}
+}
+
+func getBufioReaderPool(size int) *sync.Pool {
+	bufioReaderPoolLock.RLock()
+	if pool := bufioReaderPool[size]; pool != nil {
+		bufioReaderPoolLock.RUnlock()
+		return pool
 	}
+	bufioReaderPoolLock.RUnlock()
```
achille-roussel, line 731: How about simplifying the synchronization logic to use a simple mutex? Access to this global variable should be infrequent and the critical section is short, plus `sync.RWMutex` still contends on an atomic value internally so it scales poorly with multiple cores (which I assume would be the intent of using it here).

file.go
```diff
@@ -714,6 +718,22 @@ func getBufioReader(r io.Reader) *bufio.Reader {
 func putBufioReader(rbuf *bufio.Reader) {
 	if rbuf != nil {
 		rbuf.Reset(nil)
-		bufioReaderPool.Put(rbuf)
+		getBufioReaderPool(rbuf.Size()).Put(rbuf)
```
achille-roussel, line 721: There are cases where the `bufio` package does not retain the input size, which would case the buffers to be released to a different pool than the one they were created from; for example https://cs.opensource.google/go/go/+/refs/tags/go1.19:src/bufio/bufio.go;l=53-57

There are two ways we could address: either pass the buffer size as input parameter, or return both the `*bufio.Reader` and `*sync.Pool` from `getBufioReader` so the caller has access to the pool to release the buffer to. The latter has the advantage of not requiring a second synchronization and lookup of the pool when releasing the reader.

achille-roussel (commented):
The functions get tested indirectly by all the other tests, and they are short so I'm not concerned about not having a specific unit test for those 👍

file.go
joe-elliott: No issues swapping to `sync.Mutex`, it does simplify the logic, at a likely low cost.

Do the number of cores matter? Couldn't a goroutine yield in the critical section and block other attempts to `getBufioReaderPool()` even on a single core?

Also, unsure if this matters, but we are considering a design which would likely have thousands of parquet files in a single process. We already have a hundreds.

file.go
joe-elliott: I reviewed the `bufio.Reader` logic and, as far as I can tell, it doesn't resize itself during normal operation.

Although, you are right that if you create a reader with a buffer size of < 16 it will increase the size to 16. I suppose this would mean that if someone were to open up different parquet files in the same process with very small buffer sizes they'd all be pooled together with size 16?

As is I don't think the `bufio.Reader` behavior is concerning, although who knows what future versions of Go will bring. We could do something like drop a reader on put whose size doesn't exist in the map? or panic?

achille-roussel: > Although, you are right that if you create a reader with a buffer size of < 16 it will increase the size to 16. I suppose this would mean that if someone were to open up different parquet files in the same process with very small buffer sizes they'd all be pooled together with size 16?

I believe the consequence would be that no pooling would apply, parquet-go would always look for buffers in the pool matching the buffer size exactly, then release them to the buffer pool for size 16.

I agree it's a corner case, but it highlights that the `bufio.Reader` API does not promise to respect the buffer size that it is constructed with; robust software must not rely on it as we cannot anticipate the consequences if the standard library implementation changes (which it could rightfully do).

> We could do something like drop a reader on put whose size doesn't exist in the map? or panic?

Do you have concerns about the strategy I described above?
```go
rbuf, pool := getBufioReader(bufferSize)
defer pool.Put(rbuf)
```

file.go
achille-roussel: > Do the number of cores matter? Couldn't a goroutine yield in the critical section and block other attempts to getBufioReaderPool() even on a single core?

Goroutines might be preempted even when running on a single CPU (or with GOMAXPROCS=1), so there may be contention on the mutex no matter how many cores there are, tho it would be more pronounced with large number of cores.

I do think that access to this variable will be so infrequent in comparison with the rest of the code executed by a program that it would be really unlikely that we would ever measure contention on this global mutex. If it ever becomes a problem we can figure it out then, and we'll have data to verify that the change actually improves parallelism.

file.go
joe-elliott: Pushed a commit that returns the pool from `getBufioReader()` along with the buffer. No real concerns. It did require adding another struct  field on `filePages` but that seems minor.

I'm still using the `putBufioReader()` method to place it back in the pool so that the buffer is correctly reset. I suppose I could return a `put()` function from `getBufioReader()` instead? I have no strong feelings either way.

achille-roussel: Returning a function may cause the compiler to add an extra heap allocation to capture the `*sync.Pool` value to release the object to.

I think the current form is robust and clear, let's go with it 👍

achille-roussel (approved):
Thanks for the contribution!

PR #321, opened 2022-08-26
Serialize time.Time as a timestamp

### LINKED ISSUE #266: How to serialize time.Time
I'm attempting to serialize time objects, I thought I would be able to do that as either a timestamp or a date.

Here is what I tried:

```
type Data struct {
	PropertyKeyID                 int64     `csv:"PropertyKey_ID" parquet:"PropertyKey_ID,snappy"`
	Dealid                              int64     `csv:"Deal_id" parquet:"Deal_id,snappy"`
	Propertyid                        int64     `csv:"Property_id" parquet:"Property_id,snappy"`
	Portfolio                           string    `csv:"Portfolio" parquet:"Portfolio,dict,snappy"`
	Propertynb                       int64     `csv:"Property_nb" parquet:"Property_nb,snappy"`
	Statustx                            string    `csv:"Status_tx" parquet:"Status_tx,dict,snappy"`
        DealUpdatedt                   time.Time `csv:"Deal_Update_dt" parquet:"Deal_Update_dt,timestamp"`
```
I get the following error:
```
panic: struct has invalid 'timestamp' parquet tag: Deal_Update_dt time.Time csv:"Deal_Update_dt" parquet:"Deal_Update_dt,timestamp"
```

How would you recommend serializing dates? Is there a way to write a custom encoder for a column to cast the time into an int32?

Thanks for any help you can give me!

lmarburger:
Serialize `time.Time` values as Parquet timestamps. The default unit is `NANOS` and can be changed using the `timestamp()` struct tag.

```go
type timeColumn struct {
	t1 time.Time
	t2 time.Time `parquet:",timestamp(millisecond)"`
}
```

Resolves #266


column_buffer_go18.go
```diff
@@ -178,6 +181,35 @@ func writeRowsFuncOfOptional(t reflect.Type, schema *Schema, path columnPath, wr
...
+		if rows.Len() == 0 {
+			return writeRows(columns, rows, levels)
+		}
+
+		for i := 0; i < rows.Len(); i++ {
+			p := (*time.Time)(rows.Index(i))
+			a := sparse.Array{}
+			if p != nil {
+				var nanos int64
+				if !p.IsZero() {
```
achille-roussel, line 199: Why is this check necessary?

column_buffer_go18.go
```diff
@@ -178,6 +181,35 @@ func writeRowsFuncOfOptional(t reflect.Type, schema *Schema, path columnPath, wr
...
+
+	return func(columns []ColumnBuffer, rows sparse.Array, levels columnLevels) error {
+		if rows.Len() == 0 {
+			return writeRows(columns, rows, levels)
+		}
+
+		for i := 0; i < rows.Len(); i++ {
+			p := (*time.Time)(rows.Index(i))
+			a := sparse.Array{}
+			if p != nil {
```
achille-roussel, line 197: I don't think `p` can ever be `nil` in this context.

column_buffer_go18.go
```diff
@@ -178,6 +181,35 @@ func writeRowsFuncOfOptional(t reflect.Type, schema *Schema, path columnPath, wr
...
+	elemSize := uintptr(t.Size())
+	writeRows := writeRowsFuncOf(t, schema, path)
+
+	return func(columns []ColumnBuffer, rows sparse.Array, levels columnLevels) error {
+		if rows.Len() == 0 {
+			return writeRows(columns, rows, levels)
+		}
+
+		for i := 0; i < rows.Len(); i++ {
+			p := (*time.Time)(rows.Index(i))
```
achille-roussel, line 195: Would it be worth adding a specialization of `sparse.TimeArray` like we do for other types to improve type safety?

column_buffer_go18.go
abraithwaite: When working on my PR, it was hard for me to know when zero and nil checks are necessary versus when they aren't.  How can we tell whether or not they'd be necessary given the function signatures for the `funcOf` patterns?

achille-roussel: I guess Go doesn't help much here, generally I would say that pointers are never intended to be `nil` except when they are paired with an error in the return values of a function (in which case we usually test the error). If that's not the case it should be documented on exported APIs.

Panics for accessing nil pointers seem like a good enough safeguard then.

In this particular case, the `sparse.(*Array).Index` method returns a pointer to the element at the given index, so it cannot ever be nil.

value.go
```diff
@@ -633,7 +640,13 @@ func assignValue(dst reflect.Value, src Value) error {
 			dst.SetUint(uint64(v))
 			return nil
 		default:
-			val = reflect.ValueOf(v)
+			switch dst.Type() {
+			case reflect.TypeOf(time.Time{}):
+				t := time.Unix(0, v).UTC()
```
abraithwaite, line 645: Nice, I did not know you could do this!

internal/quick/quick.go
```diff
@@ -70,6 +71,13 @@ func (c *Config) Check(f interface{}) error {
 type MakeValueFunc func(reflect.Value, *rand.Rand)
 
 func MakeValueFuncOf(t reflect.Type) MakeValueFunc {
+	switch t {
+	case reflect.TypeOf(time.Time{}):
+		return func(v reflect.Value, r *rand.Rand) {
+			v.Set(reflect.ValueOf(time.Unix(0, r.Int63()).UTC()))
```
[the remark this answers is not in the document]
achille-roussel: the `internal/quick` package is used to generate test cases.

value.go
achille-roussel, line 645: I don't know if we can make the assumption that `v` will be a number of nanoseconds here. Even if as a first step we only support encoding as nanosecond values, the decoding path shouldn't assume that all parquet files out there are meant to use nanoseconds only. We would have to scale the value based on the underlying unit of the timestamp type to properly interpret the integer.

column_buffer_go18.go
lmarburger: Good catch. Would that code then look something like this?

```go
ts := rows.TimeArray()
for i := 0; i < ts.Len(); i++ {
	t := ts.Index(i)
	val := reflect.ValueOf(t.UnixNano())
	a := makeArray(unsafecast.PointerOfValue(val), 1, elemSize)
	...
}
```

wolfeidau:
Is there any plans to merge this at some point?

Can i help with contributions?

lmarburger:
I believe this change is blocked on #365. I don't think logical type information is being written to the parquet column information which causes times to be deserializable only as `int64`. I'm looking into it.

type.go
```diff
@@ -1722,11 +1722,45 @@ func (t *timestampType) Decode(dst encoding.Values, src []byte, enc encoding.Enc
 }
 
 func (t *timestampType) AssignValue(dst reflect.Value, src Value) error {
-	return Int64Type.AssignValue(dst, src)
+	switch dst.Type() {
+	case reflect.TypeOf(time.Time{}):
+		// TODO: Need to figure out the source unit rather than assuming nanos
```
achille-roussel, line 1727: The source unit is the unit of `t` isn't it?

value.go
```diff
@@ -197,10 +199,30 @@ func ValueOf(v interface{}) Value {
 		panic("cannot create parquet value from go value of type " + t.String())
 	}
 
-	return makeValue(k, reflect.ValueOf(v))
+	return makeValue(k, nil, reflect.ValueOf(v))
```
achille-roussel, line 202: Should `ValueOf` be modified to handle the case where it's called with a `time.Time`?

type.go
lmarburger: Indeed it is 👍

value.go
lmarburger: 👍 Added in f13bb96

achille-roussel (approved):
🚢

PR #328, opened 2022-08-31
Close pages in the doc example.

cyriltovena:
I had to look into the code if pages needs to be close. I think adding it to the doc makes it clear.

### DIFF
--- a/README.md
+++ b/README.md
@@ -402,6 +402,7 @@ mechanisms to read column values:
 
 ```go
 pages := column.Pages()
+defer pages.Close()
 
 for {
     p, err := pages.ReadPage()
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -402,6 +402,7 @@ mechanisms to read column values:
 
 ```go
 pages := column.Pages()
+defer pages.Close()
```
abraithwaite, line 405: This can error, so maybe something like this?

```suggestion
defer checkErr(pages.Close())
```

We don't need to define what `checkErr` does for the docs, but it maybe helpful to remind folks to check `io.Closer` errors.

cyriltovena: Yep

abraithwaite (approved):
Thanks for the contribution!

PR #329, opened 2022-08-31
don't copy byte-array values

achille-roussel:
This PR is a follow up to https://github.com/segmentio/parquet-go/pull/265 which addresses the performance regression that I had introduced to address the original bug.

I left documentation in the code, at a high level the fix is to never read more than one page per column in each call to `ReadRows`, so we don't invalidate pointers written to the output `[]Row` values.

Contributes to https://github.com/segmentio/parquet-go/issues/226
```
name                                                 old time/op  new time/op  delta
MergeFiles/BYTE_ARRAY/groups=2,rows=20000            51.4µs ± 4%  11.8µs ± 2%   -77.04%  (p=0.000 n=10+10)
MergeFiles/FIXED_LEN_BYTE_ARRAY/groups=2,rows=20000  32.9µs ± 2%   7.6µs ± 1%   -76.92%  (p=0.000 n=10+10)

name                                                 old row/s    new row/s    delta
MergeFiles/BYTE_ARRAY/groups=2,rows=20000             18.8M ± 4%   82.0M ± 2%  +335.29%  (p=0.000 n=10+10)
MergeFiles/FIXED_LEN_BYTE_ARRAY/groups=2,rows=20000   29.4M ± 2%  127.6M ± 1%  +333.30%  (p=0.000 n=10+10)
```

### DIFF
--- a/row_group.go
+++ b/row_group.go
@@ -378,10 +378,6 @@ func (r *rowGroupRows) ReadRows(rows []Row) (int, error) {
 	// the columns' page, the rows that were already read during the ReadRows
 	// call would be invalidated, and might reference memory locations that have
 	// been reused due to pooling of page buffers.
-	//
-	// The application is expected to handle the case where ReadRows returns
-	// less rows that requested and no error, by looking at the first returned
-	// value from ReadRows, which is the number of rows that were read.
 	numRows := int64(len(rows))
 
 	for i := range r.columns {
--- a/row_group.go
+++ b/row_group.go
@@ -252,30 +252,42 @@ func NewRowGroupRowReader(rowGroup RowGroup) Rows {
 
 type rowGroupRows struct {
 	rowGroup RowGroup
-	columns  []columnChunkReader
-	seek     int64
+	buffers  []Value
+	readers  []asyncPages
+	columns  []columnChunkRows
 	inited   bool
 	closed   bool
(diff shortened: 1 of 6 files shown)


row_group.go
```diff
@@ -303,112 +329,114 @@ func (r *rowGroupRows) Close() error {
...
+	// remaining in the current page of each column. This is necessary because
+	// the pointers exposed to the returned rows need to remain valid until the
+	// next call to ReadRows, SeekToRow, Reset, or Close. If we release one of
+	// the columns' page, the rows that were already read during the ReadRows
+	// call would be invalidated, and might reference memory locations that have
+	// been reused due to pooling of page buffers.
+	//
+	// The application is expected to handle the case where ReadRows returns
+	// less rows that requested and no error, by looking at the first returned
+	// value from ReadRows, which is the number of rows that were read.
```
abraithwaite, line 384: I like these comments, they're very clear!  Should we add/move this second paragraph to the `func (r *rowGroupRows) ReadRows` godoc though?

Also, minor grammar fix: "less rows that requested" -> "less rows *than* requested".

abraithwaite (approved):
Looks great!  One minor comment, but otherwise LGTM!

achille-roussel:
Thanks for the review!

PR #331, opened 2022-08-31
Panic on schemas containing nested lists without a list tag

### LINKED ISSUE #327: Infinite loop in repeatedPageValues.ReadValues
Run this test to reproduce:
```
func TestLoop(t *testing.T) {
	type testType struct {
		ListOfLists [][]int
	}
	instance := testType{ListOfLists: [][]int{{1, 3, 5}, nil, {2, 4, 6}}}
	buf := parquet.NewGenericBuffer[testType]()
	_, err := buf.Write([]testType{instance})
	require.NoError(t, err)

	rg := parquet.RowGroup(buf)
	chunks := rg.ColumnChunks()
	require.Equal(t, 1, len(chunks))
	pages := chunks[0].Pages()
	p, err := pages.ReadPage()
	require.NoError(t, err)

	vals := make([]parquet.Value, p.NumValues())
	_, _ = p.Values().ReadValues(vals)
}
```

It looks like the infinite loop is caused because this check always evaluates to false:
https://github.com/segmentio/parquet-go/blob/b03c0471ebb0eadd7fc54f41fccd58dafca80eee/page_values.go#L39

As soon as the `nil` is removed from the `testType` instantiation, the test completes successfully.

abraithwaite:
Fixes #327

### DIFF
--- a/page_values.go
+++ b/page_values.go
@@ -75,7 +75,7 @@ func (r *repeatedPageValues) ReadValues(values []Value) (n int, err error) {
 		// page size AND the current element's definitionLevel is not the
 		// maxDefinitionLevel (this is a null value), Create the zero values to be
 		// returned in this run
-		for n < len(values) && r.offset < len(definitionLevels) && definitionLevels[r.offset] < maxDefinitionLevel {
+		for n < len(values) && r.offset < len(definitionLevels) && definitionLevels[r.offset] != maxDefinitionLevel {
 			values[n] = Value{
 				repetitionLevel: repetitionLevels[r.offset],
 				definitionLevel: definitionLevels[r.offset],
--- a/page_values.go
+++ b/page_values.go
@@ -68,8 +68,17 @@ func (r *repeatedPageValues) ReadValues(values []Value) (n int, err error) {
 	repetitionLevels := r.page.repetitionLevels
 	columnIndex := ^int16(r.page.Column())
 
-	for n < len(values) && r.offset < len(definitionLevels) {
-		for n < len(values) && r.offset < len(definitionLevels) && definitionLevels[r.offset] != maxDefinitionLevel {
+	outBufLen := len(values)
+	pageLen := len(definitionLevels)
+
+	// while we haven't exceeded the output buffer and we haven't exceeded the page size
+	for n < outBufLen && r.offset < pageLen {
+
+		// while we haven't exceeded the output buffer and we haven't exceeded the
+		// page size AND the current element's definitionLevel is not the
+		// maxDefinitionLevel (this is a null value), Create the zero values to be
+		// returned in this run
+		for n < outBufLen && r.offset < pageLen && definitionLevels[r.offset] < maxDefinitionLevel {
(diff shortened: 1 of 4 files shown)


page_values.go
```diff
@@ -68,8 +68,17 @@ func (r *repeatedPageValues) ReadValues(values []Value) (n int, err error) {
...
+	pageLen := len(definitionLevels)
+
+	// while we haven't exceeded the output buffer and we haven't exceeded the page size
+	for n < outBufLen && r.offset < pageLen {
+
+		// while we haven't exceeded the output buffer and we haven't exceeded the
+		// page size AND the current element's definitionLevel is not the
+		// maxDefinitionLevel (this is a null value), Create the zero values to be
+		// returned in this run
+		for n < outBufLen && r.offset < pageLen && definitionLevels[r.offset] < maxDefinitionLevel {
```
abraithwaite, line 81: Reset variable names, keep comments.

schema.go
```diff
@@ -846,6 +846,15 @@ func makeNodeOf(t reflect.Type, name string, tag []string) Node {
 		node = List(node)
 	}
 
+	if node.Repeated() && !list {
+		elemKind := node.GoType().Elem().Kind()
+		switch elemKind {
+		case reflect.Slice:
+			panic(fmt.Sprintf("unhandled nested slice without list tag: %v", elemKind))
```
abraithwaite, line 853: elemKind isn't helpful here, it'll always be "slice"

page_values.go
```diff
@@ -68,7 +68,13 @@ func (r *repeatedPageValues) ReadValues(values []Value) (n int, err error) {
 	repetitionLevels := r.page.repetitionLevels
 	columnIndex := ^int16(r.page.Column())
 
+	// while we haven't exceeded the output buffer and we haven't exceeded the page size
 	for n < len(values) && r.offset < len(definitionLevels) {
+
+		// while we haven't exceeded the output buffer and we haven't exceeded the
+		// page size AND the current element's definitionLevel is not the
+		// maxDefinitionLevel (this is a null value), Create the zero values to be
+		// returned in this run
```
achille-roussel, line 77:
```suggestion
		// While we haven't exceeded the output buffer and we haven't exceeded the
		// page size AND the current element's definitionLevel is not the
		// maxDefinitionLevel (this is a null value), Create the zero values to be
		// returned in this run.
```

page_values.go
```diff
@@ -68,7 +68,13 @@ func (r *repeatedPageValues) ReadValues(values []Value) (n int, err error) {
 	repetitionLevels := r.page.repetitionLevels
 	columnIndex := ^int16(r.page.Column())
 
+	// while we haven't exceeded the output buffer and we haven't exceeded the page size
```
achille-roussel, line 71:
```suggestion
	// While we haven't exceeded the output buffer and we haven't exceeded the page size.
```

achille-roussel (approved):
I made a few formatting suggestions, looking good otherwise!

PR #333, opened 2022-09-01
README: Fix type references in README for reading pages

abraithwaite:
~The types (`parquet.<Type>Page`) are not currently exported.  Is the intent to export them?  If so we can toss this PR and make that change instead.~

Updated references in readme to correct reader interface types.

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,11 +410,11 @@ for {
     }
 
     switch page := p.Values().(type) {
-    case parquet.Int32Page:
+    case parquet.Int32Reader:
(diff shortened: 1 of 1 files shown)


achille-roussel (commented):
I think the docs did not get updated after some types were renamed; this is what they are supposed to refer to https://pkg.go.dev/github.com/segmentio/parquet-go#Int32Reader

abraithwaite:
Cool, I'll update to that then!

achille-roussel (approved):

Issue #332, opened 2022-09-01
Slice out of bounds panics while using dictionaries


joe-elliott:
During larger queries we see a number of different panics related to slice out of bounds in dictionaries.

Example 1:
```panic: index out of bounds

goroutine 4690 [running]:
github.com/segmentio/parquet-go.errno.check(...)
	/home/joe/dev/joe-elliott/tempo/vendor/github.com/segmentio/parquet-go/errors.go:65
github.com/segmentio/parquet-go.(*uint64Dictionary).bounds(0x1649ce5337840846?, {0xc083856000?, 0xc1c29b2e112ce4d?, 0x0?})
	/home/joe/dev/joe-elliott/tempo/vendor/github.com/segmentio/parquet-go/dictionary_amd64.go:134 +0x99
github.com/segmentio/parquet-go.(*uint64Dictionary).Bounds(0xc090e5c2c0, {0xc083856000?, 0xa0bde1c12?, 0x3632e40?})
	/home/joe/dev/joe-elliott/tempo/vendor/github.com/segmentio/parquet-go/dictionary.go:1072 +0x45
github.com/segmentio/parquet-go.(*indexedPage).Bounds(0xc07096a7e0)
	/home/joe/dev/joe-elliott/tempo/vendor/github.com/segmentio/parquet-go/dictionary.go:1293 +0x9c
github.com/grafana/tempo/pkg/parquetquery.(*IntBetweenPredicate).KeepPage(0xc06e347b80, {0x257c0f0?, 0xc07096a810?})
	/home/joe/dev/joe-elliott/tempo/pkg/parquetquery/predicates.go:176 +0x6b
github.com/grafana/tempo/pkg/parquetquery.(*InstrumentedPredicate).KeepPage(0xc06e599140, {0x257c0f0?, 0xc07096a810?})
	/home/joe/dev/joe-elliott/tempo/pkg/parquetquery/predicates.go:208 +0x43
github.com/grafana/tempo/pkg/parquetquery.(*ColumnIterator).iterate.func3({0x257b3e0, 0xc06b23be60}, {0x2565588, 0xc06e59a990}, 0xc0600cdf58, 0xc0600cded0, 0xc060c9b400, {0xc06ea4e000, 0x3e8, 0x3e8}, ...)
	/home/joe/dev/joe-elliott/tempo/pkg/parquetquery/iters.go:346 +0x24b
[...]

PR #335, opened 2022-09-01
add parquet-go module version to CreatedBy file metadata

achille-roussel:
This PR adds the parquet-go module version to the default value of the `CreatedBy` field in file metadata. This information will be valuable in the future to link back parquet files to the version of parquet-go that was used to generate them.

### DIFF
--- a/config.go
+++ b/config.go
@@ -2,13 +2,14 @@ package parquet
 
 import (
 	"fmt"
+	"runtime/debug"
 	"strings"
+	"sync"
 
 	"github.com/segmentio/parquet-go/compress"
 )
 
 const (
-	DefaultCreatedBy            = "github.com/segmentio/parquet-go"
 	DefaultColumnIndexSizeLimit = 16
 	DefaultColumnBufferCapacity = 16 * 1024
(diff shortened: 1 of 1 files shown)


achille-roussel:
For context, this would have been used to have when investigating issues like https://github.com/segmentio/parquet-go/issues/332

abraithwaite (approved):
Nice!

config.go
```diff
@@ -19,6 +20,50 @@ const (
...
+
+var (
+	defaultCreatedByInfo string
+	defaultCreatedByOnce sync.Once
+)
+
+func defaultCreatedBy() string {
+	defaultCreatedByOnce.Do(func() {
+		createdBy := parquetGoModulePath
+		build, ok := debug.ReadBuildInfo()
```
kevinburkesegment, line 35: Is this available on 1.17? Ok by me if it is not

kevinburkesegment (approved):

config.go
achille-roussel: It was added in Go 1.12!

PR #337, opened 2022-09-06
close rowGroupCursors when Popped due to io.EOF in mergedRowGroupRows

asubiotto:
This commit fixes a potential goroutine leak. mergeRowGroupRows cursors, when
initialized, might spawn a goroutine (e.g. asyncPages.readPages). Previously,
when a cursor returned io.EOF, this cursor was popped from the min-heap but
never closed, resulting in leaked goroutines.

cc @achille-roussel @Pryz

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -120,7 +120,12 @@ func (r *mergedRowGroupRows) ReadRows(rows []Row) (n int, err error) {
 				r.err = err
 				return n, err
 			}
+			c := r.cursors[0]
 			heap.Pop(r)
+			if err := c.close(); err != nil {
+				r.err = err
+				return n, err
+			}
 		} else {
 			heap.Fix(r, 0)
 		}
--- a/merge_test.go
+++ b/merge_test.go
@@ -17,6 +17,75 @@ const (
 	rowsPerGroup = benchmarkNumRows
 )
 
+type wrappedRowGroup struct {
+	parquet.RowGroup
+	rowsCallback func(parquet.Rows) parquet.Rows
+}
+
+func (r wrappedRowGroup) Rows() parquet.Rows {
+	return r.rowsCallback(r.RowGroup.Rows())
+}
+
+type wrappedRows struct {
+	parquet.Rows
+	closed bool
+}
+
+func (r *wrappedRows) Close() error {
+	r.closed = true
+	return r.Rows.Close()
+}
+
+func TestMergeRowGroupsCursorsAreClosed(t *testing.T) {
+
+	type model struct {
+		A int
+	}
+
+	schema := parquet.SchemaOf(model{})
+	options := []parquet.RowGroupOption{
+		parquet.SortingColumns(
+			parquet.Ascending(schema.Columns()[0]...),
+		),
+	}
+
+	prng := rand.New(rand.NewSource(0))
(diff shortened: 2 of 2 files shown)


asubiotto:
Any pointers on where to add a regression test?

merge.go
```diff
@@ -120,7 +120,12 @@ func (r *mergedRowGroupRows) ReadRows(rows []Row) (n int, err error) {
 				r.err = err
 				return n, err
 			}
-			heap.Pop(r)
+			if c, ok := heap.Pop(r).(rowGroupCursor); ok {
```
achille-roussel, line 123: Maybe we could avoid the type assertion by taking this approach?
```go
evict := r.cursors[0]
heap.Pop(r)
if err := evict.close(); err != nil {
    ...
}
```

achille-roussel (commented):
Thanks for the contribution!

Regarding regression tests, we could create a wrapper of `parquet.RowGroup` which overrides the `Rows()` method and returns a wrapped `parquet.Rows` which counts the number of `Close` calls; then we can verify that all open rows are closed?

merge.go
asubiotto: I want to push back on that. I think the problem with that is that it introduces subtle bugs since it's not explicitly clear that we're closing the cursor that was popped from the heap. Your suggested approach implies the caller knows which cursor was popped from the heap (i.e. how the internal `Pop` implementation works). In fact, the snippet is wrong since the cursor that is actually popped from the heap is the last one in the list (i.e. `s/r.cursors[0]/r.cursors[len(r.cursors-1)]`). I think the type assertion is worth it for more explicit code but also happy to change if you disagree.

achille-roussel: The heap interface actually documents it:
https://pkg.go.dev/container/heap
> The minimum element in the tree is the root, at index 0.

https://pkg.go.dev/container/heap#Pop
> Pop removes and returns the minimum element (according to Less) from the heap.

There is some ambiguity, since the heap index are not required to map to the slice indexes, but they do in our implementation, so we actually have strong guarantees that the code will remain correct across future Go versions.

I tend to prefer optimizing for compile time type checks, which also has the benefit of removing the runtime cost of interface assertion. I agree it's a trade off, if you want to keep it as-is I'm OK with it.

joe-elliott:
We are possibly seeing this in a rollout today. We have ~100s of orphaned goroutines parked here:

https://github.com/segmentio/parquet-go/blob/main/page.go#L222

The goroutines do seem to be slowly falling off. Currently still investigating.

image

Is that consistent with this issue?

achille-roussel:
If you are using the Merge* functions it would be related. 

Alternatively you might be missing a Close call on Rows instances in the code using parquet-go.

joe-elliott:
Yup, this was my bad. Saw some new behavior on a rollout and was concerned, but this is attributable to other changes in the rollout. I'll leave my original post only for continuity of the conversation :P

asubiotto:
Updated and added test.

asubiotto:
@joe-elliott there are some finalizers set up to close resources on GC (including the goroutines leaked by the behavior this PR fixes). I found it helpful to disable those and run a leak detector to track down this leak.

achille-roussel (approved):
Thank you for the fix!

PR #354, opened 2022-09-29
Feature: Leveled buffers

joe-elliott:
This PR introduces leveled buffering to the pools in buffer.go. Each pool internally is divided into 16 different pools. The smallest pool is guaranteed to have buffers in size from 1024->2047. The next is 2048->4095, etc. When a buffer is requested the appropriate subpool is used to guarantee that the buffer will be large enough. This prevents all buffers in the pools to trend towards max requested size. The largest pool contains buffers of 32MB and larger. This is the one sub pool that does not guarantee that it will return a buffer of the appropriate size and a resize may be necessary.

We have seen a total heap reduction of ~25% in the components we tested.

Also, fixed a bug where `resize()` would drop a buffer.

### DIFF
--- a/buffer.go
+++ b/buffer.go
@@ -1,7 +1,6 @@
 package parquet
 
 import (
-	"fmt"
 	"math/bits"
 	"sort"
 	"sync"
@@ -396,7 +395,6 @@ func levelledPoolIndex(sz int) int {
 	i = 32 - bits.LeadingZeros32(uint32(i)) // log2
 	if i >= totalPools {
 		i = totalPools - 1
-		fmt.Println("this happened", sz)
 	}
 	if i < 0 {
 		i = 0
--- a/buffer.go
+++ b/buffer.go
@@ -307,16 +307,6 @@ func (b *buffer) unref() {
 	}
 }
 
-func (b *buffer) clone() (clone *buffer) {
-	if b.pool != nil {
-		clone = b.pool.get(len(b.data))
-	} else {
-		clone = &buffer{refc: 1}
-	}
-	clone.data = append(clone.data, b.data...)
-	return clone
-}
-
 // slice of sync.pools is used for levelled buffering.
 // the table below shows the pools used for different buffer sizes. the first range of
 // values are the sizes for which the pool will be used on get. the second range
--- a/buffer.go
+++ b/buffer.go
@@ -348,11 +348,11 @@ func (b *buffer) clone() (clone *buffer) {
 // ...
 // [14] -> 16MB
 // [15] -> 32MB
-const totalPools = 16
+const numPoolBuckets = 16
 const basePoolIncrement = 1024
 
 type bufferPool struct {
-	pool [totalPools]sync.Pool
+	pool [numPoolBuckets]sync.Pool
 }
 
 // get returns a buffer from the levelled buffer pool. sz is used to choose the appropriate pool
@@ -394,8 +394,8 @@ func (p *bufferPool) put(b *buffer) {
 func levelledPoolIndex(sz int) int {
 	i := sz / basePoolIncrement
 	i = 32 - bits.LeadingZeros32(uint32(i)) // log2
-	if i >= totalPools {
-		i = totalPools - 1
+	if i >= numPoolBuckets {
+		i = numPoolBuckets - 1
 	}
 	if i < 0 {
 		i = 0
--- a/buffer.go
+++ b/buffer.go
@@ -307,29 +307,6 @@ func (b *buffer) unref() {
 	}
 }
 
-func (b *buffer) reset() {
-	b.data = b.data[:0]
-}
-
-func (b *buffer) resize(size int) {
-	if cap(b.data) < size {
-		const pageSize = 4096
-		minSize := 2 * cap(b.data)
-		bufferSize := ((size + (pageSize - 1)) / pageSize) * pageSize
-		if bufferSize < minSize {
-			bufferSize = minSize
-		}
-		// put old byte slice back in the pool
-		buffer := newBuffer(b.data)
-		buffer.pool = b.pool
-		buffer.unref() // unref will put it back in the pool
-
-		b.data = make([]byte, size, bufferSize)
-	} else {
-		b.data = b.data[:size]
-	}
-}
-
 func (b *buffer) clone() (clone *buffer) {
 	if b.pool != nil {
 		clone = b.pool.get(len(b.data))
@@ -341,13 +318,17 @@ func (b *buffer) clone() (clone *buffer) {
 }
 
 // slice of sync.pools is used for levelled buffering.
-// min size of buffer in each pool is basePoolIncrement*2^i
+// the table below shows the pools used for different buffer sizes. the first range of
+// values are the sizes for which the pool will be used on get. the second range
+// are the sizes of the buffers that are placed into the pool on put. when allocating a new buffer
+// from a given pool we always choose the min of the put range to guarantee that all gets
+// will have an adequately sized buffer.
 //
-// [0] -> 1024
-// [1] -> 2048
+// [pool] : <get range>  : <put range>
+// [0]    : 0    -> 1023 : 1024 -> 2047
+// [1]    : 1024 -> 2047 : 2048 -> 4095
+// [2]    : 2048 -> 4095 : 4096 -> 8191
 // ...
-// [14] -> 16MB
-// [15] -> 32MB
 const numPoolBuckets = 16
 const basePoolIncrement = 1024
 
@@ -362,16 +343,17 @@ func (p *bufferPool) get(sz int) *buffer {
 	if b == nil {
 		// align size to the pool
 		poolSize := basePoolIncrement << i
-		if poolSize > sz {
-			sz = poolSize
+		if sz > poolSize { // this can occur when the buffer requested is larger than the largest pool
+			poolSize = sz
 		}
 		b = &buffer{
-			data: make([]byte, 0, sz),
+			data: make([]byte, 0, poolSize),
 			pool: p,
 		}
-	} else {
-		b.reset()
 	}
+	// the above guarantees that the buffer is at least sz bytes long
+	b.data = b.data[:sz]
+
 	b.ref()
 	return b
 }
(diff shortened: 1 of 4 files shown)


abraithwaite:
For my own understanding, this is essentially creating room by allowing smaller buffers to still be created without claiming ownership of a big buffer even when there's very large buffers being allocated for I/O.  Is that understanding correct?

Also, without anything to prevent resize of buffers by the owner when it's claimed from a pool (not sure we'd want to do this), it's possible that a buffer would be taken from a smaller pool and returned to a larger pool, yeah?  I presume this wouldn't be a big issue in practice with the size hints, but I'm curious how we're thinking about that from a `bufferPool` consumer perspective.

Very cool though, thanks!  I think we should let Achille take a look before merging, but overall I think it looks great.

joe-elliott:
> For my own understanding, this is essentially creating room by allowing smaller buffers to still be created without claiming ownership of a big buffer even when there's very large buffers being allocated for I/O. Is that understanding correct?

Yup, this will choose more appropriately sized buffers when they are requested. This prevents all buffers trending toward the largest size. It also significantly reduces allocs b/c previously it would be very common to receive a smaller buffer then requested.

> Also, without anything to prevent resize of buffers by the owner when it's claimed from a pool (not sure we'd want to do this), it's possible that a buffer would be taken from a smaller pool and returned to a larger pool, yeah?

Yup. I wanted to build something with as small of an impact to the codebase as possible. So all the existing code that pulls from the pool and then attempts to resize still continues working. This implementation still does have one case in which this is necessary: when you are dealing in buffers > 32MB.

Going to mark this draft real fast. There is one more change I want to make and test. As written if you request a 2047 byte buffer it will return one of exactly this size and then on `put()` it will place that buffer in the 1024 byte bucket. This is a bit wasteful. I'm going to change these lines:

> if sz < basePoolIncrement {
> 	sz = basePoolIncrement
> }

To align to the next pool up.

column.go
```diff
@@ -628,15 +628,15 @@ func (c *Column) decodeDataPage(header DataPageHeader, numValues int, repetition
...
 		defer vbuf.unref()
 		vbuf.resize(int(pageType.EstimateSize(numValues)))
 		pageValues = vbuf.data
 	}
 	if pageKind == ByteArray {
-		obuf = pageOffsetsBufferPool.get()
+		obuf = pageOffsetsBufferPool.get(4 * (numValues + 1))
 		defer obuf.unref()
-		obuf.resize(4 * (numValues + 1))
+		obuf.resize(4 * (numValues + 1)) // jpe - wut?
```
achille-roussel, line 639: If you intend to keep this comment, do you mind expanding it a bit to carry more context? It's a bit hard to grasp to meaning as-is.

buffer.go
```diff
@@ -326,22 +332,43 @@ func (b *buffer) resize(size int) {
...
 
+// slice of sync.pools is used for levelled buffering.
+// min size of buffer in each pool is basePoolIncrement*2^i
+//
+// [0] -> 1024
+// [1] -> 2048
+// ...
+// [14] -> 16MB
+// [15] -> 32MB
+const totalPools = 16
```
achille-roussel, line 351: Maybe name something like `numPoolBuckets` or similar? `totalPools` makes it sound like this is the global number of pools in the program, despite being a value local to each buffer pool.

column.go
```diff
@@ -488,7 +488,7 @@ func schemaRepetitionTypeOf(s *format.SchemaElement) format.FieldRepetitionType
 }
 
 func (c *Column) decompress(compressedPageData []byte, uncompressedPageSize int32) (page *buffer, err error) {
-	page = uncompressedPageBufferPool.get()
+	page = uncompressedPageBufferPool.get(int(uncompressedPageSize))
 	if uncompressedPageSize > 0 {
 		page.resize(int(uncompressedPageSize))
 	}
```
achille-roussel, line 494: I'm curious what you'd think about potentially removing the `resize` method? We are now passing the size to `get` which could set the buffer size accordingly already?

column.go
```diff
@@ -707,7 +707,7 @@ func decodeLevelsV2(enc encoding.Encoding, numValues int, data []byte, length in
 }
 
 func decodeLevels(enc encoding.Encoding, numValues int, data []byte) (levels *buffer, err error) {
-	levels = levelsBufferPool.get()
+	levels = levelsBufferPool.get(len(data) * 8)
```
achille-roussel, line 710: It seems like maybe this might be a bit too conservative? Nested columns would result in repetition or definition levels that are encoded with more than one bit, would it make this call to `get` always allocate a new buffer then?

buffer.go
```diff
@@ -353,7 +380,27 @@ func (p *bufferPool) put(b *buffer) {
 	if b.pool != p {
 		panic("BUG: buffer returned to a different pool than the one it was allocated from")
 	}
-	p.pool.Put(b)
+	// if this slice is somehow less then our min pool size, just drop it
+	sz := cap(b.data)
+	if sz < basePoolIncrement {
+		return
+	}
+	i := levelledPoolIndex(sz / 2) // divide by 2 to put the buffer in the level below so it will always be large enough
```
achille-roussel, line 388: In cases where the program did not grow the byte slice with `append`, would this division result in always returning the buffer to the `sync.Pool` _below_ the one it was allocated from?

buffer.go
```diff
@@ -318,6 +319,11 @@ func (b *buffer) resize(size int) {
 		if bufferSize < minSize {
 			bufferSize = minSize
 		}
+		// put old byte slice back in the pool
+		buffer := newBuffer(b.data)
+		buffer.pool = b.pool
+		buffer.unref() // unref will put it back in the pool
+
```
achille-roussel, line 326: Did you notice that this was significantly improving memory reuse? I'm asking to make sure it's worth the complexity (e.g. instead of leaving the GC take care of reclaiming memory), as there are other cases where we append to the buffers byte slices for example which result in GC work anyways.

column.go
joe-elliott: apologies, i leave my initials "jpe" as reminders to go back and review code. this shouldn't be here.

column.go
joe-elliott: my assumption was that the code used `resize()` method aside from immediately after pulling from the pool, but that doesn't seem to be the case. a quick attempt to remove this caused a failing test. i'll give it a real try early next week.

column.go
joe-elliott: yeah, i presume that the next line `enc.DecodeLevels` would have to resize the buffer accordingly. any guesses on a constant that makes more sense? 16? 32?

buffer.go
```diff
@@ -326,22 +332,43 @@ func (b *buffer) resize(size int) {
 
 func (b *buffer) clone() (clone *buffer) {
```
abraithwaite, line 333: Unrelated to the content of this PR, I was curious where this was used, but it's not used anywhere so should we delete it @achille-roussel ?

buffer.go
joe-elliott: It's my intention for it to go back into the same pool.

If, for instance, you request a 2500 byte buffer it would belong to pool index 2.
pool 0: 0 -> 1023
pool 1: 1024 -> 2047
pool 2: 2048 -> 4095

these lines would align it to the beginning of the next pool. so the code actually allocates a 4096 byte buffer.
```
		poolSize := basePoolIncrement << i
		if poolSize > sz {
			sz = poolSize
		}
```

then on put 4096 / 2 = 2048 which goes back in the pool index 2.

buffer.go
joe-elliott: I did not test this change independently. Since we're planning on just removing `resize()` then this should go away.

joe-elliott:
I know it's not standard practice in this repo to write internal tests, but I'd love to test these changes directly :)

achille-roussel:
We have some internal tests in places where it makes sense, feel free to add more if you think they are useful 👍

column.go
joe-elliott: oh, one other question. if `get()` is supposed to start doing the work of resize do we see value in keeping the code to align slices to 4096 bytes?

achille-roussel: 🤔 aligning the allocations to the size class seems like the right approach to me, do you have any concerns?

column.go
joe-elliott: ok, i feel silly. The parameter `numValues` tells us exactly how big the buffer needs to be. fixed.

joe-elliott:
I think I've addressed all review comments:

- unused methods `resize()`, `clone()` and `reset()` removed
- comments added to `bufferPool` to show how the pools are used to help clarify behavior
- `numValues` used insetad of `len(data) * 8`. This is showing better memory performance in practice.
- added some internal tests

Thanks for reviewing!

buffer.go
```diff
@@ -306,45 +307,47 @@ func (b *buffer) unref() {
...
-}
+// bufferPool holds a slice of sync.pools used for levelled buffering.
+// the table below shows the pools used for different buffer sizes when both getting
+// and putting a buffer. when allocating a new buffer from a given pool we always choose the
+// min of the put range to guarantee that all gets will have an adequately sized buffer.
+//
+// [pool] : <get range>  : <put range>  : <alloc size>
+// [0]    : 0    -> 1023 : 1024 -> 2047 : 1024
+// [1]    : 1024 -> 2047 : 2048 -> 4095 : 2048
+// [2]    : 2048 -> 4095 : 4096 -> 8191 : 4096
```
abraithwaite, line 318: This comment helps a bunch, thanks!

buffer.go
```diff
@@ -306,45 +307,47 @@ func (b *buffer) unref() {
...
+		b = &buffer{
+			data: make([]byte, 0, poolSize),
+			pool: p,
+		}
 	}
+	// if the buffer comes from the largest pool it may not be big enough
+	if cap(b.data) < sz {
+		p.pool[i].Put(b)
+		b = &buffer{
+			data: make([]byte, 0, sz),
```
achille-roussel, line 346: Should we align `sz` on the bucket upper bound here?

joe-elliott: technically this shouldn't be necessary. the only time that `cap(b.data) < sz` could be true is if sz > 32MB and the pool happens to pull a buffer that isn't large enough. however, since this behavior is spread across two functions and is far from obvious i have no issues adding code directly here to support that if you'd prefer.

achille-roussel: OK, let's try it out as-is and see if we ever measure issues with this approach 👍

achille-roussel (approved):

Issue #349, opened 2022-09-27
Question about append write parquet file.


ycyang-26:
Can this project write new data to an existing parquet file in AWS S3?

ycyang-26:
Thanks for your answer! Looking forward to adding this function soon~

ycyang-26:
> Hello @ycyang-26, thanks for asking this question!
> 
> Parquet files have an immutable structure, there is no native support for appending data in the Parquet specification. That being said, there are two ways to emulate append operations on Parquet files:
> 
> * A file may contain multiple row groups, which are written somewhat back-to-back to a files. A program could technically append a row group and rewrite the file footer to include the new metadata, and the resulting file would still be a valid parquet file. Now things get tricky when other file sections are present, like page indexes or bloom filters. Currently there is no support in parquet-go for this, but we could consider adding it.
> * An alternative solution which may be a bit more compute intensive would be a rewrite a new file which contains the content of the old file plus the extra data. In this case, the application may also have more control over the file layout (e.g. merging the extra data in an existing row group). This requires extra memory/storage to hold the new file, plus compute time to decode and re-encode the file data. In some cases, this is the desirable approach, and it can be achieved using parquet-go, here is a code example:
> 
> ```go
> r := parquet.NewGenericReader[Row](inputFile)
> w := parquet.NewGenericWriter[Row](outputFile)
> 
> _, err := parquet.CopyRows(w, r)
> if err != nil {
>     ...
> }
> 
> // append new rows to the writer + close when done
> ...
> ```
> 
> Let me know if the answer was useful!

Hi,
Thanks about the answer. I'm wondering that how to use the example code? I want to read several parquet files from s3 and concat to a large data. Then write the data to a new parquet file in s3. I'm confused about the example code.

ycyang-26:
I want to merge small files to a large file which data size equals to 1GB. So I try to load rows from parquet files and then append to a slice. But the row data size seems too large and throws error because out of memory. So is possible to merge small data to a large size by your following example code?

achille-roussel:
I'll leave all the parts to read/write S3 out of the example because this is not specific to parquet-go.

It is possible to append rows to a parquet writer in multiple calls to `Write`, `WriteRows`, or multiple `CopyRows`. To concatenate multiple parquet files into one: create a writer to generate the output file, then open each input file and copy their rows to the writer using `CopyRows`.

If maintaining row ordering is required, using `MergeRowGroups` could also be an option here.

[...]

PR #356, opened 2022-09-29
infer the given encoding based on the provided encodings

thorfour:
Proposed solution to https://github.com/segmentio/parquet-go/issues/349 to infer the given schema encoding from the column encodings array

### DIFF
--- a/column.go
+++ b/column.go
@@ -331,9 +331,19 @@ func (cl *columnLoader) open(file *File, path []string) (*Column, error) {
 			// each page of the column should iterate through the pages and read
 			// the page headers to determine which compression and encodings are
 			// applied.
-			for _, encoding := range c.chunks[0].MetaData.Encoding {
-				c.encoding = LookupEncoding(encoding)
-				break
+			switch encodings := c.chunks[0].MetaData.Encoding; len(encodings) {
+			case 1:
+				c.encoding = LookupEncoding(encodings[0])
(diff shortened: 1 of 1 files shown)


column.go
```diff
@@ -331,9 +331,19 @@ func (cl *columnLoader) open(file *File, path []string) (*Column, error) {
...
+				c.encoding = LookupEncoding(encodings[0])
+			default: // If RLE, Plain, and an additional encoding are present, the additional encoding is the given one
+				for _, encoding := range encodings {
+					if encoding != format.Plain && encoding != format.RLE {
+						c.encoding = LookupEncoding(encoding)
+						break
+					}
+				}
+				if c.encoding == nil {
+					c.encoding = LookupEncoding(format.Plain)
```
achille-roussel, line 345: Is it possible that we might be wrongly selecting the encoding here? For example if the list of encodings was `[RLE, RLE]` (not sure if it's valid Parquet, but we would be deducting `PLAIN` even tho it wasn't present).

thorfour: Based on the way the writer wraps encodings this is correct (since something like RLE, RLE can't happen) but your point makes sense that maybe a different library wrote that.

thorfour: Updated, to use the default behavior in the nominal case

achille-roussel (approved):
Looking great! Thanks for the fix!

PR #357, opened 2022-09-30
internal/debug: add build directive to disable runtime.SetFinalizer

asubiotto:
This commit adds a build directive to disable runtime.SetFinalizer if the
`debug` tag is specified. The use-case is to disable cleaning up resources so
that users can detect leaked goroutines more easily, since Finalizers are
currently used to close asyncPages and rowGroupRows if not done so before.

cc @achille-roussel

### DIFF
--- a/internal/debug/finalizer_off.go
+++ b/internal/debug/finalizer_off.go
@@ -1,0 +1,6 @@
+//go:build debug
+
+package debug
+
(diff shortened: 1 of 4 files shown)


achille-roussel (commented):
What do you think about moving the code to the existing internal/debug package?

achille-roussel (approved):

PR #363, opened 2022-10-05
fix issue 362

### LINKED ISSUE #362: Memory Corruption when using parquet.ReadFile multiple times in a row
First of all, thanks for much for open sourcing this project, I found it at exactly the right time as the previous package I was using had massive memory problems and I didn't want to have to move to using java for messing with parquet files

I was noticing some weird problems while using parquet.ReadFile[T] (affects parquets.Read[T]) as well. Basically I was testing a function of have to modify the contents of a parquet file. So I was using `parquet.ReadFile[any]` and `parquet.Read[any]` to be able to easily get the results to compare in my tests. And I was seeing some very strange things around corruption of `[]byte` values. I was able to track this down to a second call corrupting the data from the first call.

My guess is that this has to do with a lot of the unsafe logic that is going on in the package and the fact that we no longer have a reference to the `parquet.GenericReader[T]` that was used to read the rows. My tests were failing intermittently so I'm wondering if it is based on whether or not garbage collection happened in between. If I am right about this, it seems like this package level functions aren't actually safe to use.

The tests below `TestParquetReadFromGenericReaders` and `TestParquetReadFileWithCopyingResults` don't seem to ever fail, but `TestParquetReadFile` will fail sometimes.

```go
package dmsparquet

import (
	"io"
	"os"
	"testing"

	"github.com/mitchellh/copystructure"
	"github.com/segmentio/parquet-go"
	"github.com/stretchr/testify/assert"
	"github.com/stretchr/testify/require"
)

func TestParquetReadFromGenericReaders(t *testing.T) {
	path := "../testdata/dms_test_table_LOAD00000001.parquet"
	fp, err := os.Open(path)
	require.NoError(t, err)
	defer fp.Close()

	r1 := parquet.NewGenericReader[any](fp)
	rows1 := make([]any, r1.NumRows())
	_, err = r1.Read(rows1)
	if err != nil && err != io.EOF {
		require.NoError(t, err)
	}

	r2 := parquet.NewGenericReader[any](fp)
	rows2 := make([]any, r2.NumRows())
	_, err = r2.Read(rows2)
	if err != nil && err != io.EOF {
		require.NoError(t, err)
	}
}

func TestParquetReadFileWithCopyingResults(t *testing.T) {

	rows1, err := parquet.ReadFile[any]("../testdata/dms_test_table_LOAD00000001.parquet")
	require.NoError(t, err)

	for i, row := range rows1 {
		rows1[i] = copystructure.Must(copystructure.Copy(row))
	}

	rows2, err := parquet.ReadFile[any]("../testdata/dms_test_table_LOAD00000001.parquet")
	require.NoError(t, err)

	assert.ElementsMatch(t, rows1, rows2)
}

func TestParquetReadFile(t *testing.T) {

	rows1, err := parquet.ReadFile[any]("../testdata/dms_test_table_LOAD00000001.parquet")
	require.NoError(t, err)

	rows2, err := parquet.ReadFile[any]("../testdata/dms_test_table_LOAD00000001.parquet")
	require.NoError(t, err)

	assert.ElementsMatch(t, rows1, rows2)
}
```
[dms_test_table_LOAD00000001.parquet.tgz](https://github.com/segmentio/parquet-go/files/9712181/dms_test_table_LOAD00000001.parquet.tgz)

achille-roussel:
Fixes #362


value.go
```diff
@@ -695,7 +695,7 @@ func assignValue(dst reflect.Value, src Value) error {
 				return nil
 			}
 		default:
-			val = reflect.ValueOf(v)
+			val = reflect.ValueOf(copyBytes(v))
```
achille-roussel, line 698: This is the fix.

kevinburkesegment (approved):

abraithwaite (approved):

PR #364, opened 2022-10-06
add concise search example test

abraithwaite:
A common workflow for consumers will be searching for items contained in their parquet files.

This demonstrates how how one can do it with the library today.

### DIFF
--- a/parquet_go18_test.go
+++ b/parquet_go18_test.go
@@ -121,6 +121,67 @@ func ExampleWrite_any() {
 	// map["FirstName":"R2" "LastName":"D2"]
 }
 
+func ExampleSearch() {
+	type Row struct{ FirstName, LastName string }
+
+	buf := new(bytes.Buffer)
+	// The column being searched must be sorted for search to work.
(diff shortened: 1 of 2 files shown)


parquet_go18_test.go
```diff
@@ -121,6 +121,67 @@ func ExampleWrite_any() {
 	// map["FirstName":"R2" "LastName":"D2"]
 }
 
+func ExampleSearch() {
+	type Row struct{ FirstName, LastName string }
+
+	buf := new(bytes.Buffer)
+	// The column being searched must be sorted for search to work.
```
kevinburkesegment, line 128: My first question reading this comment is, how is the column sorted? First name? Last name? Is this specified somewhere?

kevinburkesegment (approved):

parquet_go18_test.go
abraithwaite: I had the same question, so I should probably add some clarity here once I know the proper way! 😅

PR #371, opened 2022-10-12
Collapse Buffer Pools to 1

joe-elliott:
Previously we reduced overall memory consumption of parquet-go by [leveling our buffer pools](https://github.com/segmentio/parquet-go/pull/354). This PR collapses all buffer pools into 1 to reduce memory consumption even further. 

I believe that they were broken out into individual pools to help consolidate buffers of similar sizes in different pools. However, with the leveling change sharing pools should reduce overall consumption. In practice we are seeing roughly 5% reduction in working set on the most heavily impacted processes.

### DIFF
--- a/buffer.go
+++ b/buffer.go
@@ -380,11 +380,7 @@ func levelledPoolIndex(sz int) int {
 }
 
 var (
-	levelsBufferPool           bufferPool
-	compressedPageBufferPool   bufferPool
-	uncompressedPageBufferPool bufferPool
-	pageOffsetsBufferPool      bufferPool
-	pageValuesBufferPool       [8]bufferPool
+	bytesPool bufferPool
 )
 
 type bufferedPage struct {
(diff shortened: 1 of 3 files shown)


buffer.go
```diff
@@ -380,11 +380,7 @@ func levelledPoolIndex(sz int) int {
 }
 
 var (
-	levelsBufferPool           bufferPool
-	compressedPageBufferPool   bufferPool
-	uncompressedPageBufferPool bufferPool
-	pageOffsetsBufferPool      bufferPool
-	pageValuesBufferPool       [8]bufferPool
+	bytesPool bufferPool
```
achille-roussel, line 383: At first I felt like this name communicated something like "a pool of one-byte values". 

Have you considered an alternative name? Maybe `buffers`?

joe-elliott: works for me. changed!

achille-roussel (approved):

Issue #368, opened 2022-10-10
Occasional `bad use of unsafe.Pointer?` panics during GC


joe-elliott:
This panic has returned. Previously it was reported here:

https://github.com/segmentio/parquet-go/issues/309

We are seeing a similar error in a different component. It is rare, but sometimes happens in groups likely due to write or read patterns.

```
runtime: marked free object in span 0x7f73ee799028, elemsize=24 freeindex=239 (bad use of unsafe.Pointer? try -d=checkptr)
0xc146cd2000 alloc unmarked
0xc146cd2018 alloc unmarked
0xc146cd2030 alloc marked  
0xc146cd2048 alloc unmarked
0xc146cd2060 alloc unmarked
0xc146cd2078 alloc unmarked
0xc146cd2090 alloc unmarked
0xc146cd20a8 alloc unmarked
...
0xc146cd3668 free  marked   zombie
0x000000c146cd3668:  0x7265732e746f6f72  0x6d616e2e65636976 
0x000000c146cd3678:  0x00656e6364333065 
...
fatal error: found pointer to free object
```

Some examples:
![image](https://user-images.githubusercontent.com/2272392/194914606-960d9a10-2483-47d3-a549-bcf18765d623.png)

achille-roussel:
Hello @joe-elliott, thanks for reporting!

Would you happen to know if the issue occurs when your application is built with with `-tags purego`?

joe-elliott:
I do not. This is only currently occurring in a single environment that is difficult to experiment in. Going to spend the next few days trying to repro in a way that will make it easier to explore the issue.

joe-elliott:
Ok, I have a setup that can reproduce this extremely reliably. It includes private data that I cannot share, but I can definitely build with different options or make other code changes to help hunt it down.

> Would you happen to know if the issue occurs when your application is built with with -tags purego?
I am unable to reproduce the issue when building with `-tags purego`. I can reproduce it 100% of the time on a normal build within 3 or 4 queries.

Attached is the pqrs schema dump for the data.parquet file that I can reproduce this issue on. Next I will begin making changes to our code. If I can identify what calls into `parquet-go` cause the issue hopefully that will help track it down.

[out.txt](https://github.com/segmentio/parquet-go/files/9780569/out.txt)

The code that is regularly tripping the issue begins here: 
https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet/block_findtracebyid.go#L73

[...]

PR #373, opened 2022-10-14
Test file for issue #368

joe-elliott:
As discussed out of band, this file paired with the following test trip #368 100% of the time.

The Trace object referenced is here: https://github.com/grafana/tempo/blob/main/tempodb/encoding/vparquet/schema.go#L157

```
func TestIterFile(t *testing.T) {
	for i := 0; i < 5; i++ {
		f, err := os.OpenFile("file.parquet")
		require.NoError(t, err)

		info, err := f.Stat()
		require.NoError(t, err)

		pf, err := parquet.OpenFile(f, info.Size())
		require.NoError(t, err)

		reader := parquet.NewGenericReader[*Trace](pf)

		trs := make([]*Trace, 1)
		for {
			_, err := reader.Read(trs)
			if err != nil {
				break
			}
		}

		reader.Close()
		f.Close()
	}
}
```


achille-roussel:
Closing in favor of #376 which includes the file and test. Thanks for providing this repro!

PR #380, opened 2022-10-19
fix issue 370

### LINKED ISSUE #370: How to return a Page to the pool?
Hi, Grafana Tempo uses the low-level API for scanning columns, including ColumnChunk, PageReader, etc. There have been performance improvements to the library recently to introduce reference-counted buffers and I think there might be a gap where a `Page` (backed by `bufferedPage`) isn't returned to the pool, nor is there a way to do it manually.  It's possible that I missed the proper usage of these low-level structs, so this issue is both a possible finding and an ask for clarification.

Our [loop](https://github.com/grafana/tempo/blob/main/pkg/parquetquery/iters.go#L327) is basically:

```go
for _, rowGroup := range rowGroups {
    pages := rowGroup.ColumnChunks[n].Pages()
    defer pages.Close()

    for {
        page := pages.ReadPage()
        ...
```

It looks like new memory is allocated for every call to ReadPage(), and not reused from the internal pool, via a `bufferedPage` with new refs in [column.decodeDataPage](https://github.com/segmentio/parquet-go/blob/main/column.go#L675).   I didn't see where the page is [unref'd](https://github.com/segmentio/parquet-go/blob/main/buffer.go#L413) to return the buffers to the pool.

**Thoughts**

Should this be happening automatically, as the semantics of `ReadPage` mention that the given page is only valid until the next call?    Another possibility is to introduce `Page.Close()` which unrefs the buffers and allows downstream applications finer control over the lifetime.

achille-roussel:
Fixes #370 

The solution implemented is the one discussed on the issue, please let me know if you have any concerns!

### DIFF
--- a/buffer.go
+++ b/buffer.go
@@ -406,18 +406,24 @@ func (p *bufferedPage) Slice(i, j int64) Page {
 	}
 }
 
-func unref(page Page) {
-	if p, _ := page.(*bufferedPage); p != nil {
-		bufferUnref(p.values)
-		bufferUnref(p.offsets)
-		bufferUnref(p.definitionLevels)
-		bufferUnref(p.repetitionLevels)
+func (p *bufferedPage) Retain() {
+	bufferRef(p.values)
+	bufferRef(p.offsets)
+	bufferRef(p.definitionLevels)
+	bufferRef(p.repetitionLevels)
+}
 
-		p.values = nil
-		p.offsets = nil
-		p.definitionLevels = nil
-		p.repetitionLevels = nil
-	}
+func (p *bufferedPage) Release() {
+	bufferUnref(p.values)
+	bufferUnref(p.offsets)
+	bufferUnref(p.definitionLevels)
+	bufferUnref(p.repetitionLevels)
+
+	p.Page = nil
+	p.values = nil
+	p.offsets = nil
+	p.definitionLevels = nil
+	p.repetitionLevels = nil
 }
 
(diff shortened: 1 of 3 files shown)


buffer.go
```diff
@@ -431,3 +437,49 @@ func bufferUnref(buf *buffer) {
...
 	}
 }
+
+// Retain is a helper function to increment the reference counter of pages
+// backed by memory which can be granularly managed by the application.
+//
+// Usage of this function is optional and with Release, is intended to allow
+// finer grain memory management in the application. Most programs should be
+// able to rely on automated memory management provided by the Go garbage
+// collector instead.
```
kevinburkesegment, line 447: I would maybe give an example of when you would _want_ to use this

buffer.go
```diff
@@ -431,3 +437,49 @@ func bufferUnref(buf *buffer) {
...
+func Retain(page Page) {
+	if p, _ := page.(retainable); p != nil {
+		p.Retain()
+	}
+}
+
+// Release is a helper function to decrement the reference counter of pages
+// backed by memory which can be granularly managed by the application.
+//
+// Usage of this is optional and with Retain, is intended to allow finer grain
```
kevinburkesegment, line 460:
```suggestion
// Usage of this is optional and with Retain, is intended to allow finer grained
```

kevinburkesegment (approved):

abraithwaite (approved):

PR #387, opened 2022-10-25
Refactor assigning Parquet values to Go values

lmarburger:
As part of exploring #266, it was identified that the code around deserializing Parquet values into Go values should be refactored to make supporting other types like `time.Time` possible. This is the first change which refactors the monolithic `assignValue()` function into an `AssignValue()` function implemented on each Parquet type.


type.go
```diff
@@ -1290,6 +1463,9 @@ func (t *timeType) useInt32() bool {
 }
 
 func (t *timeType) useInt64() bool {
+	// TODO: Write test covering this change. Not even sure if this is correct.
+	//       Nothing in this struct refers to nanos.
+	//return t.Unit.Micros != nil || t.Unit.Nanos != nil
```
achille-roussel, line 1468: I'm not sure I understand why this behavior needed to be changed.

type.go
```diff
@@ -1118,6 +1277,11 @@ func (t *jsonType) Decode(dst encoding.Values, src []byte, enc encoding.Encoding
 	return encoding.DecodeByteArray(dst, src, enc)
 }
 
+func (t *jsonType) AssignValue(dst reflect.Value, src Value) error {
+	// TODO: Is this correct?
```
achille-roussel, line 1281: It's correct.

type.go
lmarburger: Oops. This slipped through from another change I was working through. Removed in d532fa2

type.go
```diff
@@ -194,6 +201,8 @@ var (
 	FloatType     Type = floatType{}
 	DoubleType    Type = doubleType{}
 	ByteArrayType Type = byteArrayType{}
+	// TODO: Rename this
+	BE128Type Type = fixedLenByteArrayType{length: 16}
```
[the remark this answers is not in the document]
lmarburger: That works for me. 👍 1e7baf6

type.go
```diff
@@ -238,6 +247,15 @@ var convertedTypes = [...]deprecated.ConvertedType{
...
 }
 
+func defaultAssignValue(val reflect.Value, dst reflect.Value, src Value) error {
+	if !val.IsValid() || !val.Type().AssignableTo(dst.Type()) {
+		return fmt.Errorf("cannot assign parquet value of type %s to go value of type %s", src.Kind().String(), dst.Type())
+	}
+
+	dst.Set(val)
+	return nil
+}
```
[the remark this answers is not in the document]
lmarburger: Oh in that case we can inline this. I thought we wanted to preserve the validation checking so this function mainly existed to consolidate that and the error returned if it failed. Fixed in 81e90c4

row.go
```diff
@@ -739,10 +739,18 @@ func reconstructFuncOfGroup(columnIndex int16, node Node) (int16, reconstructFun
...
 		if !row.startsWith(columnIndex) {
 			return row, fmt.Errorf("no values found in parquet row for column %d", columnIndex)
 		}
-		return row[1:], assignValue(value, row[0])
+
+		//// TODO: This behavior isn't captured here
+		//if src.IsNull() {
+		//	dst.Set(reflect.Zero(dst.Type()))
+		//	return nil
+		//}
```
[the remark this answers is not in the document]
lmarburger: None at all. I wanted to get your opinion on if it belonged here. e31a1d4

type.go
```diff
@@ -278,6 +285,18 @@ func (t booleanType) Decode(dst encoding.Values, src []byte, enc encoding.Encodi
...
+	v := src.Boolean()
+	switch dst.Kind() {
+	case reflect.Bool:
+		dst.SetBool(v)
+		return nil
+	}
+
+	dst.Set(reflect.ValueOf(v))
+	return nil
+}
```
achille-roussel, line 298: Maybe rewrite these methods to have only one `return nil` instead of having to return in each case?
```suggestion
	v := src.Boolean()
	switch dst.Kind() {
	case reflect.Bool:
		dst.SetBool(v)
	default:
		dst.Set(reflect.ValueOf(v))
	}
	return nil
}
```

type.go
```diff
@@ -519,6 +598,24 @@ func (t byteArrayType) Decode(dst encoding.Values, src []byte, enc encoding.Enco
...
+	v := src.ByteArray()
+	switch dst.Kind() {
+	case reflect.String:
+		dst.SetString(string(v))
+		return nil
+	case reflect.Slice:
+		if dst.Type().Elem().Kind() == reflect.Uint8 {
+			dst.SetBytes(copyBytes(v))
+			return nil
+		}
```
achille-roussel, line 611: How about not checking so it panics if the underlying type is not compatible, which would also be the same behavior if we reached the code below?
```suggestion
		dst.SetBytes(copyBytes(v))
		return nil
```

type.go
```diff
@@ -571,6 +668,33 @@ func (t fixedLenByteArrayType) Decode(dst encoding.Values, src []byte, enc encod
...
+			// a more efficient call to copy.
+			d := unsafe.Slice((*byte)(unsafecast.PointerOfValue(dst)), len(v))
+			copy(d, v)
+			return nil
+		}
+	case reflect.Slice:
+		if dst.Type().Elem().Kind() == reflect.Uint8 {
+			dst.SetBytes(copyBytes(v))
+			return nil
+		}
```
achille-roussel, line 690: Same suggestion as above:
```suggestion
		dst.SetBytes(copyBytes(v))
		return nil
```

achille-roussel (commented):
This is shaping up! 🙌

type.go
lmarburger: Good idea. Updated all relevant switches in 6119f39

achille-roussel (approved):
Looking good 👍

row.go
```diff
@@ -739,10 +739,21 @@ func reconstructFuncOfGroup(columnIndex int16, node Node) (int16, reconstructFun
...
 		if !row.startsWith(columnIndex) {
 			return row, fmt.Errorf("no values found in parquet row for column %d", columnIndex)
 		}
-		return row[1:], assignValue(value, row[0])
+
+		var err error
+		src := row[0]
+
+		if src.IsNull() {
+			value.Set(reflect.Zero(value.Type()))
```
achille-roussel, line 752: I wonder if this would be better implemented by being moved to each `AssignValue` method: in most cases we go through the specialized code path (e.g. `SetInt`) where a null parquet value would result in the zero value and have the same effect but would be more efficient since the code wouldn't have to instantiate the zero `reflect.Value`.

We are also bypassing type checking in this branch, where we don't validate that the destination Go value was actually assignable from parquet values of the parquet type.

lmarburger: 🤔 That's a good idea. Looks like the only function on `Value` that doesn't appreciate being called on a null value is `ByteArray()` (and `Int96()` by extension). All the others return a zero value so the implementation can be ignorant of null values. How does c6efc31 look?

achille-roussel: That's surprising, on a null value the ptr+len will be nil and zero 
https://github.com/segmentio/parquet-go/blob/main/value.go#L405 which should still be a valid byte slice?
```go
func (v Value) ByteArray() []byte { return unsafe.Slice(v.ptr, int(v.u64)) }
```

achille-roussel: Let's make a special case for Int96 only then 👍

achille-roussel:
:shipit:

row.go
lmarburger: Done in 3381746

PR #388, opened 2022-10-25
search: add comments to search.go

abraithwaite:
I was going through this function to check my understanding of page indexes while thinking about granules in TraceDB.

I believe that granules in other databases like clickhouse and frostdb are the equivlant of column pages in parquet.  The page index is the sparse index over the pages (granules) wherein the data being searched for is contained.

This check on the binary search mostly confirms that suspicion, if I'm not mistaken.  Please correct me if I'm misunderstaneding something.

Also, as part of this, I found what is possibly an optimization to avoid skipping ahead a page, by comparing the current value to the max in the current page, instead of letting the binary search check the next page, then falling through to the final if-block.

It may or may not be a bug also, if, for example, the following happens:

you have pages:

    [1,2,3]
    [3,4,5]
    [6,7,8]

Searching for 3 in this case would not return the first page and instead return the second, missing values of 3 in the first page.

I'll have to look at writing a test for this, but the current search test will pass creating this scenario without altering the test somewhat (since it's only looking for presence, not the expected index)

### DIFF
--- a/search.go
+++ b/search.go
@@ -121,8 +121,7 @@ func binarySearch(index ColumnIndex, value Value, cmp func(Value, Value) int) in
 		}
 	}
 
-	// If value may exist somewhere in the column
-	// e.g. we got to idx 0 or idx n-1
+	// last page check, if it wasn't explicitly found above
 	if curIdx < n {
 
 		// check current nextIdx for value
--- a/search.go
+++ b/search.go
@@ -81,18 +81,17 @@ func binarySearch(index ColumnIndex, value Value, cmp func(Value, Value) int) in
 	// while there's at least one more page to check
 	for (topIdx - curIdx) > 1 {
 
-		// nextIdx is halfway between curIdx and topIdx
+		// nextIdx is set to halfway between curIdx and topIdx
 		nextIdx := ((topIdx - curIdx) / 2) + curIdx
 
 		smallerThanMin := cmp(value, index.MinValue(nextIdx))
-		greaterThanMax := cmp(value, index.MaxValue(nextIdx))
 
 		switch {
-		// search below this page
+		// search below pages[nextIdx]
 		case smallerThanMin < 0:
 			topIdx = nextIdx
-		// search above this page
-		case greaterThanMax > 0:
+		// search pages[nextIdx] and above
+		case smallerThanMin > 0:
 			curIdx = nextIdx
 		case smallerThanMin == 0:
 			// this case is hit when winValue == value of nextIdx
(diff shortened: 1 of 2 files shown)


search.go
```diff
@@ -75,33 +75,67 @@ func Find(index ColumnIndex, value Value, cmp func(Value, Value) int) int {
...
+
+			// if the page proceeding this has a maxValue matching the value we're
+			// searching, continue the search.
+			// otherwise, we can return early
+			//
+			// cases covered by else block
+			// if cmp(value, index.MaxValue(nextIdx-1)) < 0: the value is only in this page
+			// if cmp(value, index.MaxValue(nextIdx-1)) > 0: we've got a sorting problem with overlapping pages
+			if nextIdx-1 > curIdx && cmp(value, index.MaxValue(nextIdx-1)) == 0 {
+				topIdx = nextIdx
```
abraithwaite, line 114: It maybe more efficient to do a linear search from this on out, but I don't want to assume that outright (e.g. null values at the beginning of a column could have many pages worth of data, so still worth binary searching it?).

achille-roussel: I don't think we can assume that nulls will be at the beginning, they might be placed at the end depending on how the column is configured.

achille-roussel (commented):
If I'm understanding correctly the current version of the algorithm returns the index of the last page with min value greater or equal to the value being searched.

I'm curious whether an alternative approach could be to do a linear search backwards from that index after the binary search? It might keep the code simpler and avoid doing two calls to the comparison function per iteration?

Edit: since linear search will have a O(N) worst case we could also do a second binary search on [0:index] looking for the first page where the max value is greater or equal to the value we are looking for?

search.go
abraithwaite: Sure, it was just an example argument against doing a linear search in reverse.

achille-roussel:
> My understanding is that the current algorithm returns early if it finds a page with minValue == searchValue

OK, my mistake.

> How is this different from what's implemented in this PR?

The difference would be that we would only do one comparison per iteration: first loop compares min values, second compares max values.

The change you are proposing is good, I just want to explore whether there are opportunities for simplification 👍

achille-roussel (approved):

abraithwaite:
> The difference would be that we would only do one comparison per iteration: first loop compares min values, second compares max values.

Gotcha. I did think about this given the existing code was only doing one bounds check (on `index.MinValue`) a bit and couldn't figure out a way to handle this kind of case:

```
search: 5
[1,2,3]
[4,5,6]
[7,8,9]
```

Only checking Min on the first loop would skip to page [7,8,9] and would return a bad result (not in columnChunk), if I'm not mistaken.

abraithwaite:
Took the feedback regarding multiple checks on every iteration and implemented it, while still not regressing on the bug found previously (#389).

PR #393, opened 2022-10-27
Convert values between Parquet logical types

lmarburger:
In order to support deserializing time values from Parquet files in #266, there needs to be a way to convert between different Parquet types. For example, there might be a `TIMESTAMP(isAdjustedToUTC=true, unit=MILLIS)` value stored in a file and it needs to be deserialized into a `time.Time` which expects a number of nanoseconds. This PR adds a `ConvertValue(val Value, typ Type)` to `parquet.Type` structs which will accept the source logical type and convert it where necessary to the destination logical type.

This PR plus #387 are the last remaining changes in order to support serializing and deserializing times for #266.


type.go
```diff
@@ -188,9 +188,15 @@ type Type interface {
...
-	// Assigns a Parquet value to a Go value. Returns an error if conversion is
-	// not possible.
+	// Assigns a Parquet value to a Go value. Returns an error if assignment is
+	// not possible. It is expected that Value is using an expected logical type
+	// which can be done using ConvertValue.
 	AssignValue(dst reflect.Value, src Value) error
+
+	// Convert a Parquet value of the given logical type into a Parquet value
+	// of expected logical type. The converted Value is suitable to be passed to
+	// AssignValue.
```
achille-roussel, line 198: Could we make it more explicit that the method receiver is the type to convert _to_?

type.go
```diff
@@ -297,6 +303,10 @@ func (t booleanType) AssignValue(dst reflect.Value, src Value) error {
 	return nil
 }
 
+func (t booleanType) ConvertValue(val Value, typ Type) (Value, error) {
+	return val, nil
```
achille-roussel, line 307: How about having a generic check that the kinds match at least for cases where we don't have conversion rules yet?

```suggestion
	return val, checkTypeKindEqual(t, typ)
```

```go
func checkTypeKindEqual(to, from Type) error {
	if to.Kind() != from.Kind() {
		return fmt.Errorf("cannot convert from parquet value of type %s to %s", from, to)
	}
	return nil
}
```

achille-roussel (commented):
This is going to be such a useful addition to the library 👍

type.go
lmarburger: That definitely seems useful. Updated in e516ee8

type.go
lmarburger: 👍 Sure thing. Addressed in 8f31958

PR #399, opened 2022-11-03
Allow key value metadata to be set after writing rows

### LINKED ISSUE #397: Write key value metadata after rows
I see how to use the `KeyValueMetadata` function to configure a writer when constructing that writer.  This is useful if the key value metadata is known ahead of writing the rows.  I'm hoping to find a way to write key value metadata after writing rows.  In my case, the key value metadata includes summary information about the rows.  I'm trying to write the parquet file in a streaming manner and wanted to avoid buffering all rows to create the key value metadata summary.

I tried using `NewWriterConfig`, creating a reader with `NewGenericReader`, and then modifying the `config.KeyValueMetadata` after writing rows.  However, `NewGenericReader` creates a new config internally, so the one I am able to modify is no longer used.

Is there a way to add key value metadata after writing rows and before writing the file footer?

tschaub:
This adds a `KeyValueMetadata` method to the writer to allow key value file metadata to be updated after the writer is created.  This allows for writing metadata that might be derived from the rows without having to buffer all the rows before creating the writer.

Fixes #397.

### DIFF
--- a/writer.go
+++ b/writer.go
@@ -210,6 +210,29 @@ func (w *Writer) ReadRowsFrom(rows RowReader) (written int64, err error) {
 // The returned value will be nil if no schema has yet been configured on w.
 func (w *Writer) Schema() *Schema { return w.schema }
 
+// SetKeyValueMetadata sets a key/value pair in the Parquet file metadata.
+//
+// Keys are assumed to be unique, if the same key is repeated multiple times the
+// last value is retained. While the parquet format does not require unique keys,
+// this design decision was made to optimize for the most common use case where
+// applications leverage this extension mechanism to associate single values to
+// keys. This may create incompatibilities with other parquet libraries, or may
+// cause some key/value pairs to be lost when open parquet files written with
+// repeated keys. We can revisit this decision if it ever becomes a blocker.
+func (w *Writer) SetKeyValueMetadata(key, value string) {
+	for i, kv := range w.writer.metadata {
+		if kv.Key == key {
+			kv.Value = value
+			w.writer.metadata[i] = kv
+			return
+		}
+	}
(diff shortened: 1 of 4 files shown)


writer.go
```diff
@@ -210,6 +210,29 @@ func (w *Writer) ReadRowsFrom(rows RowReader) (written int64, err error) {
...
+// KeyValueMetadata sets a key/value pair in the Parquet file metadata.
+//
+// Keys are assumed to be unique, if the same key is repeated multiple times the
+// last value is retained. While the parquet format does not require unique keys,
+// this design decision was made to optimize for the most common use case where
+// applications leverage this extension mechanism to associate single values to
+// keys. This may create incompatibilities with other parquet libraries, or may
+// cause some key/value pairs to be lost when open parquet files written with
+// repeated keys. We can revisit this decision if it ever becomes a blocker.
+func (w *Writer) KeyValueMetadata(key, value string) {
```
achille-roussel, line 222: How about renaming to `Set*` to better communicate that the previous value will be overwritten?

We could also imagine having an `Add*` method for cases where the application needs to associated multiple values to a key.
```suggestion
func (w *Writer) SetKeyValueMetadata(key, value string) {
```

achille-roussel (commented):
Should we add these methods to the `GenericWriter` type as well?

writer.go
tschaub: Updated in the latest commit.

tschaub:
> Should we add these methods to the `GenericWriter` type as well?

Latest commit adds the method to `GenericWriter` as well.

achille-roussel (approved):
Thanks for your contributions!

PR #403, opened 2022-11-08
Run tests on main, display status of main

### LINKED ISSUE #402: Build status badge shows failing tests
I was curious if it was intentional that the tests are not run on the default branch.  As it is now, the [build status badge](https://pkg.go.dev/github.com/segmentio/parquet-go#pkg-overview) often shows failing tests since the latest action might be from any PR.

image

I understand there might be a good reason that the tests are not run on `main`.  If not, would a change like this be acceptable: https://github.com/segmentio/parquet-go/compare/main...tschaub:parquet-go:test-main ?

tschaub:
This configures the "Test" workflow to run on pull requests to `main` and on commits pushed to `main`.  Previously, this workflow was only run on pull requests, so there wasn't really a definitive indicator of whether `main` was in good shape.  

The badge displayed on the [pkg docs](https://pkg.go.dev/github.com/segmentio/parquet-go#section-readme) and readme previously showed the status of the test results for whatever pull request might have been most recently run.  The change here proposes showing the results for the latest run on the `main` branch instead.

Fixes #402.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,6 +1,11 @@
 name: Test
 on:
-- pull_request
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - '*'
 
 jobs:
   test:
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -1,6 +1,11 @@
 name: Test
 on:
-- pull_request
+  push:
+    branches:
+      - main
+  pull_request:
+    branches:
+      - main
```
tschaub, line 8: This makes it so the workflow is triggered on pull requests to `main` and on pushes to `main`.  If you want to run on pull requests to any branch, it could be updated to look like:
```yaml
on:
  push:
    branches:
      - main
  pull_request:
    branches:
      - *
```

(Or the `branches` property could be omitted altogether from the `pull_request` object.)

achille-roussel: I see, that sounds good 👍

achille-roussel: Actually I do think we want to run the workflow on any pull requests, we often split the work into multiple branches, only the top-most being derived from `main`.

tschaub: Latest commit is updated to run on pull requests to all branches.

achille-roussel (approved):
Thanks again for your contributions!

PR #417, opened 2022-11-30
Add support for synchronous page reading

joe-elliott:
This PR adds a File level config option that controls whether or not pages are read synchronously. It defaults to async to maintain backwards compatibility.

I'm not married to any of the names in this PR so feel free to suggest alternatives.

cc @asubiotto

### DIFF
--- a/buffer.go
+++ b/buffer.go
@@ -256,7 +256,7 @@ func (buf *Buffer) WriteRowGroup(rowGroup RowGroup) (int64, error) {
 //
 // The buffer and the returned reader share memory. Mutating the buffer
 // concurrently to reading rows may result in non-deterministic behavior.
-func (buf *Buffer) Rows() Rows { return &rowGroupRows{rowGroup: buf} }
+func (buf *Buffer) Rows() Rows { return newRowGroupRows(buf, PageReadModeAsync) }
 
 // bufferWriter is an adapter for Buffer which implements both RowWriter and
 // PageWriter to enable optimizations in CopyRows for types that support writing
--- a/buffer.go
+++ b/buffer.go
@@ -256,7 +256,7 @@ func (buf *Buffer) WriteRowGroup(rowGroup RowGroup) (int64, error) {
 //
 // The buffer and the returned reader share memory. Mutating the buffer
 // concurrently to reading rows may result in non-deterministic behavior.
-func (buf *Buffer) Rows() Rows { return newRowGroupRows(buf, PageReadModeAsync) }
+func (buf *Buffer) Rows() Rows { return newRowGroupRows(buf, ReadModeSync) }
 
 // bufferWriter is an adapter for Buffer which implements both RowWriter and
 // PageWriter to enable optimizations in CopyRows for types that support writing
--- a/config.go
+++ b/config.go
@@ -14,8 +14,8 @@ import (
 type ReadMode int
 
 const (
-	ReadModeAsync ReadMode = iota // ReadModeAsync reads pages asynchronously in the background.
-	ReadModeSync                  // ReadModeSync reads pages synchronously on demand.
+	ReadModeSync  ReadMode = iota // ReadModeSync reads pages synchronously on demand.
+	ReadModeAsync                 // ReadModeAsync reads pages asynchronously in the background.
 )
 
 const (
--- a/config.go
+++ b/config.go
@@ -10,6 +10,13 @@ import (
 	"github.com/segmentio/parquet-go/compress"
 )
 
+type PageReadMode int
+
+const (
+	PageReadModeAsync PageReadMode = iota
+	PageReadModeSync
+)
+
 const (
 	DefaultColumnIndexSizeLimit = 16
 	DefaultColumnBufferCapacity = 16 * 1024
@@ -83,6 +90,7 @@ type FileConfig struct {
 	SkipPageIndex    bool
 	SkipBloomFilters bool
 	ReadBufferSize   int
+	PageReadMode     PageReadMode
 }
 
 // DefaultFileConfig returns a new FileConfig value initialized with the
@@ -92,6 +100,7 @@ func DefaultFileConfig() *FileConfig {
 		SkipPageIndex:    DefaultSkipPageIndex,
 		SkipBloomFilters: DefaultSkipBloomFilters,
 		ReadBufferSize:   defaultReadBufferSize,
+		PageReadMode:     PageReadModeAsync,
 	}
 }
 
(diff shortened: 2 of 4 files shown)


asubiotto (commented):
Nice work!

buffer.go
```diff
@@ -256,7 +256,7 @@ func (buf *Buffer) WriteRowGroup(rowGroup RowGroup) (int64, error) {
 //
 // The buffer and the returned reader share memory. Mutating the buffer
 // concurrently to reading rows may result in non-deterministic behavior.
-func (buf *Buffer) Rows() Rows { return &rowGroupRows{rowGroup: buf} }
+func (buf *Buffer) Rows() Rows { return newRowGroupRows(buf, PageReadModeAsync) }
```
achille-roussel, line 259: Should we use the sync mode here since the buffer holds values in memory?

row_group.go
```diff
@@ -277,18 +278,34 @@ func (r *rowGroupRows) buffer(i int) []Value {
...
-	r.done = done
-
-	for i, column := range columns {
-		r.readers[i].init(column.Pages(), done)
+	switch r.pageReadMode {
+	case PageReadModeAsync:
+		done := make(chan struct{})
+		r.done = done
+		for i, column := range columns {
+			r.readers[i] = &asyncPages{}
```
achille-roussel, line 300: Could we allocate all `asyncPages` in a slice so the number of malloc calls doesn't grow linearly with the number of columns?

Something like this:
```go
readers := make([]asyncPages, len(columns))
for i, column := range columns {
    readers[i].init(column.Pages(), done)
    r.readers[i] = &readers[i]
}
```

config.go
```diff
@@ -10,6 +10,13 @@ import (
 	"github.com/segmentio/parquet-go/compress"
 )
 
+type PageReadMode int
+
+const (
+	PageReadModeAsync PageReadMode = iota
+	PageReadModeSync
+)
```
achille-roussel, line 18: An alternative model I had considered would be to expose this customization as an interface, for example:

```go
type PageReadMode interface {
    PageReaders(columns []ColumnChunk) ([]Pages, error)
}
```
The main upside I would imagine is that it allows the application to experiment with other strategies if needed, potentially even compose them.

We need to account for the support for asynchronous cancellation in async mode, so maybe the signature could be:
```go
type PageReadMode interface {
    PageReaders(ctx context.Context, columns []ColumnChunk) ([]Pages, error)
}
```

I'm not entirely convinced that it is a better model, but I wanted to bring it up for discussion. Let me know what you think about it.

config.go
```diff
@@ -10,6 +10,13 @@ import (
 	"github.com/segmentio/parquet-go/compress"
 )
 
+type PageReadMode int
+
+const (
+	PageReadModeAsync PageReadMode = iota
+	PageReadModeSync
```
achille-roussel, line 17: I imagine that the intent is to maintain the existing behavior by having async the default.

Thinking about what default behavior would create the least number of surprises (and maintenance work answering Github issues), I wonder if using sync as the default would be a better choice here.

config.go
brancz: I think this could be interesting! This question might already test the way people might abuse this or if it works as intended: do you imagine that this could be used to implement (custom) page caching?

buffer.go
joe-elliott: I agree with your thoughts below about defaulting to Sync because it creates the least surprises. In all places where we create `newRowGroupRows()` should we default to `PageReadModeSync`?

row_group.go
joe-elliott: done

config.go
joe-elliott: This does seem to be a question of how much flexibility we'd like to provide at this point. The idea of routing page requests to different places is tempting, but perhaps should be designed explicitly instead of added here.

I wonder if it would better handled through the Page interface? That way if you are accessing the pages of a column individually (instead of entire rows at a time) you would still have access to some kind of flexible routing behavior?

To me this feels like a specific configuration of the behavior of `rowGroupRows` then something generic.

config.go
joe-elliott: Agree. I was tempted to choose Sync, but thought backwards compat was a good choice. After we work out the questions above I can change the default to Sync.

config.go
achille-roussel: OK, your call @joe-elliott, I'm fine keeping it as-is and discussing further later on 👍

config.go
```diff
@@ -369,6 +378,15 @@ func SkipBloomFilters(skip bool) FileOption {
 	return fileOption(func(config *FileConfig) { config.SkipBloomFilters = skip })
 }
 
+// SetPageReadMode is a file configuration option which controls the way pages
+// are read. Currently the only two options are PageReadModeAsync and PageReadModeSync
+// which control whether or not pages are loaded asynchronously.
+//
+// Defaults to PageReadModeAsync.
+func SetPageReadMode(mode PageReadMode) FileOption {
```
achille-roussel, line 386: To stay consistent with other options, maybe rename?
```suggestion
func FileReadMode(mode ReadMode) FileOption {
```

config.go
```diff
@@ -10,6 +10,13 @@ import (
 	"github.com/segmentio/parquet-go/compress"
 )
 
+type PageReadMode int
```
achille-roussel, line 13: Let's add a bit of documentation on this type (and the constants as well maybe)?

joe-elliott: added some comments. let me know if you'd like anything else.

config.go
joe-elliott: done

config.go
joe-elliott: swapped default to sync

buffer.go
joe-elliott: changed to sync

config.go
joe-elliott: sounds good. let's leave it as is then and maybe discuss how to work this in more holistically.

config.go
```diff
@@ -10,6 +10,14 @@ import (
 	"github.com/segmentio/parquet-go/compress"
 )
 
+// ReadMode is an enum that is used to configure the way that a File reads pages.
+type ReadMode int
+
+const (
+	ReadModeAsync ReadMode = iota // ReadModeAsync reads pages asynchronously in the background.
```
achille-roussel, line 17: How about swapping these two constants so the zero value is sync mode?

achille-roussel (approved):
This looks great! Thanks for your contribution!

config.go
joe-elliott: done!

config.go
brancz: Works for me.

brancz (approved):

achille-roussel (approved):

PR #419, opened 2022-12-01
add parquet.Value constructors for each physical type

achille-roussel:
This PR adds constructors of `parquet.Value` from Go values representing each parquet physical type. The intent is to provide a more efficient way of constructing values than using `parquet.ValueOf` which uses reflection internally.

### DIFF
--- a/value.go
+++ b/value.go
@@ -204,7 +204,7 @@ func ValueOf(v interface{}) Value {
 	return makeValue(k, nil, reflect.ValueOf(v))
 }
 
-// BooleamValue constructs a BOOLEAN parquet value from the bool passed as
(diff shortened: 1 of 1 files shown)


value.go
```diff
@@ -204,6 +204,38 @@ func ValueOf(v interface{}) Value {
 	return makeValue(k, nil, reflect.ValueOf(v))
 }
 
+// BooleamValue constructs a BOOLEAN parquet value from the bool passed as
```
kevinburkesegment, line 207:
```suggestion
// BooleanValue constructs a BOOLEAN parquet value from the bool passed as
```

kevinburkesegment (approved):

PR #420, opened 2022-12-02
Add missing Close call in WriteRowGroup

asubiotto:
WriteRowGroup would previously not call Close on the Rows passed in to CopyRows. This could cause goroutine leaks when reading pages asynchronously.

cc @achille-roussel


achille-roussel (approved):
Thanks for the fix!

PR #422, opened 2022-12-02
Row pooling

joe-elliott:
So I can't say I expect this PR to survive exactly as is, but we are seeing significant performance improvements with this diff and I would like to get some form of row pooling into parquet-go.

- Adds a rowPool
  - This is basically a copy of the buffer pool and generics would make it easy to combine the two. Once parquet-go requires go 1.18 I think we should combine these into `slicepool[T]`
- Adds pooling mechanisms for rows
  - append is used throughout the codebase but old rows are just thrown away. `appendRow` allows us to keep the old memory and reuse it.
  - `row.Repool` and `RowFromPool` these are the changes I dislike the most, but we need the application using parquet-go to have a way to take from the pool and return to it for maximum effectiveness.
- Replaces all (that I could find) appends on rows with `appendRow`

### DIFF
--- a/column_buffer.go
+++ b/column_buffer.go
@@ -520,7 +520,8 @@ func (col *repeatedColumnBuffer) Page() Page {
 
 			if numValues > 0 {
 				if numValues > cap(col.buffer) {
-					col.buffer = make([]Value, numValues)
+					rows.put(col.buffer)
+					col.buffer = rows.get(numValues)
 				} else {
 					col.buffer = col.buffer[:numValues]
 				}
--- a/column_buffer.go
+++ b/column_buffer.go
@@ -520,7 +520,7 @@ func (col *repeatedColumnBuffer) Page() Page {
 
 			if numValues > 0 {
 				if numValues > cap(col.buffer) {
-					rows.put(col.buffer)
+					(Row)(col.buffer).Repool()
 					col.buffer = rows.get(numValues)
 				} else {
 					col.buffer = col.buffer[:numValues]
--- a/column_chunk.go
+++ b/column_chunk.go
@@ -180,7 +180,7 @@ func readRowsFuncOfLeaf(columnIndex int, repetitionDepth byte) (int, readRowsFun
 					}
 				}
 
-				rows[i] = append(rows[i], buf[col.offset])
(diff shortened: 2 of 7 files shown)


achille-roussel:
Do you have measures showing the effectiveness of this change?

I'm a bit concerned to expose memory that is pooled internally, for example in the Convert type. This is usually fertile ground for use-after-free problems since the applications has no explicit information about how it should treat the value.

joe-elliott:
> Do you have measures showing the effectiveness of this change?

We have seen a ~20% reduction in the total heap size when using this change. However, the configurables have not been tuned and I imagine we could do even better. This is a process that performs a large amount of row iteration with high variance in the length of the rows.

> I'm a bit concerned to expose memory that is pooled internally, for example in the Convert type. This is usually fertile ground for use-after-free problems since the applications has no explicit information about how it should treat the value.

Can you point to the code you're concerned about so I can better understand? It is not uncommon for a library to have methods like `Close()` or `Complete()` on objects. Calling these methods and then using the objects often gives undefined behavior. Do you see this as much different? 

I think for `parquet-go` to get to the next level of efficiency we will need to find ways to pool objects like Rows and Pages that are used both internal and external to the library. Maybe we should agree on patterns for pooling these objects that meet the expectations of users and gracefully degrade if unused.

Finally,
I don't know if this is more palatable, but I had also considered a pattern like the below. Then advanced users could add their own row pool and get the benefits while also sweating the implementation details. If no row pool is added then the default behavior is exactly like it is today.
```
type RowPool interface {
  Get(sz int) Row
  Put(r Row)
}

func AddRowPool(pool RowPool) {
...
}
```

The library could even provide some basic options if desired:
```
func ExponentialLeveledRowPool(base, buckets int) RowPool
```

joe-elliott:
Having continued to work with this change I am beginning to favor the option proposed above where the consumer of this library chooses to create and pass in their own row pool. It is likely that most applications of this library would not benefit much from this change while some, like Tempo, would benefit greatly. Our particular situation makes the ability to pool Rows very attractive. 

If the user has control over row pooling functionality then perhaps it alleviates some concerns about memory reuse? By adding their own row pool they would be making a very deliberate decision that presumably they would understand and use correctly?

PR #426, opened 2022-12-05
add parquet.RowBuffer

achille-roussel:
This PR adds a new type named `parquet.RowBuffer` which is similar to `parquet.GenericBuffer` but uses a row layout to hold data in memory (instead of a column layout).

I left documentation in the code, please take a look at let me know if anything should be changed.

### DIFF
--- a/compare.go
+++ b/compare.go
@@ -205,7 +205,7 @@ func compareRowsFuncOf(schema *Schema, sortingColumns []SortingColumn) func(Row,
 					// does not need to scan the rows looking for values with
 					// a matching column index.
 					//
-					// A second optimization consist in passing the column type
+					// A second optimization consists in passing the column type
 					// directly to the sort function instead of an intermediary
 					// closure, which removes an indirection layer and improves
 					// throughput by ~20% in BenchmarkSortRowBuffer.
--- a/compare.go
+++ b/compare.go
@@ -179,12 +179,12 @@ func lessBE128(v1, v2 *[16]byte) bool {
 
 func compareRowsFuncOf(schema *Schema, sortingColumns []SortingColumn) func(Row, Row) int {
 	compareFuncs := make([]func(Row, Row) int, 0, len(sortingColumns))
-	firstIndexOfRepeatedColumn := -1
+	direct := true
 
 	for _, column := range schema.Columns() {
 		leaf, _ := schema.Lookup(column...)
 		if leaf.MaxRepetitionLevel > 0 {
-			firstIndexOfRepeatedColumn = leaf.ColumnIndex
+			direct = false
 		}
 
 		for _, sortingColumn := range sortingColumns {
@@ -196,7 +196,7 @@ func compareRowsFuncOf(schema *Schema, sortingColumns []SortingColumn) func(Row,
 				optional := leaf.MaxDefinitionLevel > 0
 				sortFunc := (func(Row, Row) int)(nil)
 
-				if firstIndexOfRepeatedColumn < 0 && !optional {
+				if direct && !optional {
 					// This is an optimization for the common case where rows
 					// are sorted by non-optional, non-repeated columns.
 					//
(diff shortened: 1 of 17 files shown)


compare.go
```diff
@@ -6,6 +6,54 @@ import (
...
+			}
+			return +1
+		case b.IsNull():
+			return -1
+		default:
+			return cmp(a, b)
+		}
+	}
+}
+
```
achille-roussel, line 56: Moved from `search.go` as it seemed more intuitive to find these functions in `compare.go`.

search.go
```diff
@@ -1,41 +1,5 @@
...
-			}
-			return +1
-		case b.IsNull():
-			return -1
-		default:
-			return cmp(a, b)
-		}
-	}
-}
-
```
achille-roussel, line 38: Moved to `compare.go`.

search_test.go
```diff
@@ -6,28 +6,6 @@ import (
...
-}
-
-func TestCompareNullsLast(t *testing.T) {
-	cmp := parquet.CompareNullsLast(parquet.Int32Type.Compare)
-	assertCompare(t, parquet.Value{}, parquet.Value{}, cmp, 0)
-	assertCompare(t, parquet.Value{}, parquet.ValueOf(int32(0)), cmp, +1)
-	assertCompare(t, parquet.ValueOf(int32(0)), parquet.Value{}, cmp, -1)
-	assertCompare(t, parquet.ValueOf(int32(0)), parquet.ValueOf(int32(1)), cmp, -1)
-}
-
```
achille-roussel, line 30: Moved to `compare_test.go`.

compare.go
```diff
@@ -128,3 +176,135 @@ func lessBE128(v1, v2 *[16]byte) bool {
...
+				if direct && !optional {
+					// This is an optimization for the common case where rows
+					// are sorted by non-optional, non-repeated columns.
+					//
+					// The sort function can make the assumption that it will
+					// find the column value at the current column index, and
+					// does not need to scan the rows looking for values with
+					// a matching column index.
+					//
+					// A second optimization consist in passing the column type
```
kevinburkesegment, line 208:
```suggestion
					// A second optimization consists in passing the column type
```

row_buffer.go
```diff
@@ -0,0 +1,505 @@
...
+func (buf *RowBuffer[T]) copyBytes(v []byte) []byte {
+	if free := cap(buf.buffer) - len(buf.buffer); free < len(v) {
+		newCap := 2 * cap(buf.buffer)
+		if newCap == 0 {
+			newCap = defaultReadBufferSize
+		}
+		for newCap < len(v) {
+			newCap *= 2
+		}
+		buf.buffer = make([]byte, 0, newCap)
```
kevinburkesegment, line 189: Don't you need to copy the existing values here or am I nuts

row_buffer.go
```diff
@@ -0,0 +1,505 @@
...
+	numRows int
+	compare func(Row, Row) int
+}
+
+// NewRowBuffer constructs a new row buffer.
+func NewRowBuffer[T any](options ...RowGroupOption) *RowBuffer[T] {
+	config := DefaultRowGroupConfig()
+	config.Apply(options...)
+	if err := config.Validate(); err != nil {
+		panic(err)
```
kevinburkesegment, line 39: Generally I don't expect a  function called `New` to panic unless the options are _really_ obviously wrong. But I guess we should document that if so.

row_buffer.go
achille-roussel: No need, the previous rows will reference the old buffer and GC will reclaim memory when the pointers are cleared.

row_buffer.go
achille-roussel: This is the pattern we've used in this library, the application can validate the configuration prior to calling the function if it needs to validate it.

Passing invalid configuration seems like something obviously wrong to me, but that's a personal appreciation obviously :)

PR #427, opened 2022-12-05
add parquet.SortingWriter

achille-roussel:
This PR adds a new writer type named `parquet.SortingWriter` which ensures that rows written to row groups are always ordered according to the sorting columns passed as configuration.

The sorting strategy uses an in-memory buffer which gets sorted then serialized to a row group; when the writer is flushed or closed, all the row groups are merged while maintaining the global order of rows using a k-way sort.

### DIFF
--- a/sorting.go
+++ b/sorting.go
@@ -125,10 +125,7 @@ func (w *SortingWriter[T]) Flush() error {
 		return err
 	}
 
-	rows := m.Rows()
-	defer rows.Close()
-
-	if _, err := CopyRows(w.output, rows); err != nil {
+	if _, err := w.output.WriteRowGroup(m); err != nil {
 		return err
 	}
 
@@ -194,10 +191,7 @@ func (w *SortingWriter[T]) sortAndWriteBufferedRows() error {
 	defer w.rows.Reset()
 	sort.Sort(w.rows)
 
-	rows := w.rows.Rows()
-	defer rows.Close()
-
-	n, err := CopyRows(w.writer, rows)
+	n, err := w.writer.WriteRowGroup(w.rows)
 	if err != nil {
 		return err
 	}
--- a/sorting.go
+++ b/sorting.go
@@ -1,0 +1,211 @@
+//go:build go1.18
+
+package parquet
+
+import (
+	"bytes"
+	"io"
+	"sort"
+)
+
+// SortingWriter is a type similar to GenericWriter but it ensures that rows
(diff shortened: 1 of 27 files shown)


sorting.go
```diff
@@ -0,0 +1,205 @@
...
+	return w.output.Schema()
+}
+
+func (w *SortingWriter[T]) sortAndWriteBufferedRows() error {
+	if w.rows.Len() == 0 {
+		return nil
+	}
+
+	defer w.rows.Reset()
+	sort.Sort(w.rows)
```
joe-elliott, line 192: Thoughts on writing to a heap instead of sorting here?

achille-roussel: Do you mean maintaining a heap on the buffered rows as they are written?

joe-elliott: Yup, so the cost of sorting is amortized across all of the writes instead of done once when it's time to flush.

achille-roussel: The heap is not sorted tho, it seems that it means we need to load all two groups back to memory when merging since we may have to manipulate the sub heaps as they are consumed by the merge, or am I missing something?

joe-elliott: My suggestion is likely not impactful unless `maxRows` is quite large.

As far as I can tell you are buffering rows into `w.rows` in the order they are written. Then, on this line, are performing a sort on the rows so that you can flush them in order to an in memory row group. Later these are merged to write to the output file.

If, instead of a slice, `RowBufffer[T]` was backed by a heap data structure they could be retrieved in order and the cost of sorting would be amortized across all of the row writes.

However, digging deeper, the `RowBuffer` struct implements the `RowGroup` interface which includes `SeekToRow` which would be difficult to implement if all of this was backed with a heap. So, you are welcome to ignore! This was idle thought while waking up with a coffee and on deeper inspection doesn't pan out.

sorting.go
```diff
@@ -0,0 +1,205 @@
...
+package parquet
+
+import (
+	"bytes"
+	"io"
+	"sort"
+)
+
+// SortingWriter is a type similar to GenericWriter but it ensures that rows
+// are sorted according to the sorting columns configured on the writer.
```
Pryz, line 12: Somehow I thought the GenericWriter was doing that already. Otherwise what's the point of passing SortingColumns to it ? If that's not the case we might want to add something about it in the GenericWriter doc :)

Pryz (approved):

PR #430, opened 2022-12-05
add parquet.DropDuplicatedRows

achille-roussel:
Based on https://github.com/segmentio/parquet-go/pull/427, this PR adds the ability to configure the `parquet.SortingWriter` to drop duplicated rows.

As part of this PR, I also revisited the way sorting options are configured on writers and row groups.

### DIFF
--- a/row.go
+++ b/row.go
@@ -142,7 +142,9 @@ type RowWriterWithSchema interface {
 // writers passed as arguments.
 //
 // When writing rows, if any of the writers returns an error, the operation is
-// aborted and the error returned.
+// aborted and the error returned. If one of the writers did not error, but did
+// not write all the rows, the operation is aborted and io.ErrShortWrite is
+// returned.
 //
 // Rows are written sequentially to each writer in the order they are given to
 // this function.
@@ -156,9 +158,12 @@ type multiRowWriter struct{ writers []RowWriter }
 
 func (m *multiRowWriter) WriteRows(rows []Row) (int, error) {
 	for _, w := range m.writers {
-		_, err := w.WriteRows(rows)
+		n, err := w.WriteRows(rows)
 		if err != nil {
-			return 0, err
+			return n, err
+		}
+		if n != len(rows) {
+			return n, io.ErrShortWrite
 		}
 	}
 	return len(rows), nil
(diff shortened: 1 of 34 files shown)


Pryz (approved):

row.go
```diff
@@ -782,3 +784,46 @@ func reconstructFuncOfLeaf(columnIndex int16, node Node) (int16, reconstructFunc
...
+func (a *rowAllocator) copyBytes(v []byte) []byte {
+	if free := cap(a.buffer) - len(a.buffer); free < len(v) {
+		newCap := 2 * cap(a.buffer)
+		if newCap == 0 {
+			newCap = defaultReadBufferSize
+		}
+		for newCap < len(v) {
+			newCap *= 2
+		}
+		a.buffer = make([]byte, 0, newCap)
```
joe-elliott, line 819: seeing lines like this make me consider changing the `bufferPool` to be a `[]byte` pool instead of a `*buffer` pool. it would allow us to recapture slices like this. wdyt?

achille-roussel: One issue with pooling []byte is the API of sync.Pool is based on interface{}. The compiler can optimize packing pointer values into an interface without extra heap allocation but a slice is too large for that and packing it into an interface requires making an intermediary heap allocation to hold the slice value. In an effort to reduce memory pressure we end up causing heap allocations of more small objects, depending on the access pattern this can result in just as much GC work since it tends to scale based on number of objects on the heap rather than their size. 

I want to be cautious not to introduce memory pools before having a measure of their effectiveness, they always come with a non-zero maintenance cost, manual memory management is prone to more bugs than letting the GC help us out. 

Note that these allocations are expected to be amortized across large chunks of work, applications can process millions of rows and only incur a few reallocations of these buffers. Giving the GC a bit of work is OK as long as the program scales well, it buys us safety and a lower long term maintenance cost. If we have a strong signal that the pool will be more effective we can revisit then.

joe-elliott: Thank you for the detailed explanation and agree that taking on pooling should come with clear evidence of a win. Appreciate the heads up on the `interface{}` heap allocations. I did not know that was optimized away for a pointer, but not a slice.

PR #432, opened 2022-12-06
Remove parquet.Page.Clone

achille-roussel:
This PR removes the `Clone` method of the `Page` interface. This method was used as a mechanism to capture pages on columns that a writer needed to generate a bloom filter for, when we could not know ahead of time the size of the filter.

Cloning of the pages resulted in large memory utilization, which eventually contributed to reaching memory limits and causing processes to crash.

To work around this issue, we can instead trade off CPU to lower memory utilization, by generating the bloom filter from the list of pages encoded by the writer.


page_buffer.go
```diff
@@ -24,15 +24,15 @@ import (
 type PageBufferPool interface {
 	// GetPageBuffer is called when a parquet writer needs to acquires a new
```
kevinburkesegment, line 25:
```suggestion
	// GetPageBuffer is called when a parquet writer needs to acquire a new
```

kevinburkesegment (approved):

PR #436, opened 2022-12-07
fix sizing of pages

achille-roussel:
This PR contains two fixes to address excessive memory allocation.

The first issues was caused by imprecise computation of the column buffer size used when writing parquet files. I introduced a new method `parquet.Type.EstimateNumValues(size int) int` which does the opposite of `parquet.Type.EstimateSize(numValues int) int` to simplify and fix the sizing of column buffers.

The second issue was caused by having very large arrays of rows written in a single `WriteRows` call. We would only flush the column buffers after writing all the rows, causing them to grow beyond their target size. I fixed it by chunking the input to give the writer a chance to flush column buffers that fill up.

These two issues resulted in creating parquet files holding very large pages, which when reading back would require allocating equally large buffers.

### DIFF
--- a/column.go
+++ b/column.go
@@ -565,8 +565,7 @@ func (c *Column) decodeDataPageV1(header DataPageHeaderV1, page *buffer, dict Di
 		numValues -= countLevelsNotEqual(definitionLevels.data, c.maxDefinitionLevel)
 	}
 
-	page = &buffer{data: pageData}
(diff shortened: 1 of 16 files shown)


lmarburger (approved):
Thanks for pulling that magic value into `estimatedSizeOfByteArrayValues` and consolidating the logic into the various types. 😄

PR #437, opened 2022-12-07
support configuring the sorting buffer pool

achille-roussel:
This PR adds the ability to configure a buffer pool that `parquet.SortingWriter` instances use to allocate the buffer where their temporary files.

I have reused the `parquet.PageBufferPool` API for this purpose, but by default I kept the sorting buffers and column page buffers separated because they tend to be of very different sizes.

I'm open to naming suggestions, it seems `parquet.PageBufferPool` is not so representative now that we have a new use case for it.

### DIFF
--- a/buffer_pool.go
+++ b/buffer_pool.go
@@ -41,16 +41,16 @@ type BufferPool interface {
 // Go heap.
 func NewBufferPool() BufferPool { return new(memoryBufferPool) }
 
-type pageBuffer struct {
+type memoryBuffer struct {
 	data []byte
 	off  int
 }
 
-func (p *pageBuffer) Reset() {
+func (p *memoryBuffer) Reset() {
 	p.data, p.off = p.data[:0], 0
 }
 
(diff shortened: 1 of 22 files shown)


achille-roussel:
> I'm open to naming suggestions, it seems parquet.PageBufferPool is not so representative now that we have a new use case for it.

For lack of creativity I renamed `parquet.PageBufferPool` to `parquet.BufferPool`. I'm a bit concerned that it could get confusing since there is a `parquet.Buffer` type which has nothing to do with it, I'm considering moving `parquet.BufferPool` to a subpackage to avoid the confusion but haven't made up my mind on a name for the package either so for now this is the way.

kevinburkesegment (approved):

lmarburger (approved):
I agree with moving this buffer into a new package. I guess I'm not very creative today because I can't come up with a better name to differentiate the different buffers. The only thing I could come up with was renaming `parquet.Buffer` to `RowBuffer` but that already exists too. Soooo many buffers. 😆

PR #438, opened 2022-12-07
PARQUETGODEBUG=tracebuf

achille-roussel:
This PR adds facilities for debugging internal behavior of parquet-go to enable debugging capabilities at runtime. The package looks for a `PARQUETGODEBUG` environment variable to turn on and off debugging features. The first feature added here is a tracer for internal reference counted buffers.

I used this debugging facility to track down issues with the buffers reference count, which are also fixed in this PR.

### DIFF
--- a/internal/debug/debug.go
+++ b/internal/debug/debug.go
@@ -73,7 +73,13 @@ var (
 
 func init() {
 	for _, arg := range strings.Split(os.Getenv("PARQUETGODEBUG"), ",") {
-		switch k, v, _ := strings.Cut(arg, "="); k {
+		k := arg
+		v := ""
+		i := strings.IndexByte(arg, '=')
+		if i >= 0 {
+			k, v = arg[:i], arg[i+1:]
+		}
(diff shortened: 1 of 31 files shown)


Pryz (approved):

internal/debug/debug.go
```diff
@@ -63,3 +66,22 @@ func (d *ioWriter) Write(b []byte) (int, error) {
...
+	for _, arg := range strings.Split(os.Getenv("PARQUETGODEBUG"), ",") {
+		k := arg
+		v := ""
+		i := strings.IndexByte(arg, '=')
+		if i >= 0 {
+			k, v = arg[:i], arg[i+1:]
+		}
+		switch k {
+		case "tracebuf":
+			TRACEBUF, _ = strconv.Atoi(v)
```
kevinburkesegment, line 84: Panic or warn on improperly set value?

kevinburkesegment (approved):

PR #445, opened 2022-12-08
use memory allocator in byte array dictionary

achille-roussel:
Indexing BYTE_ARRAY columns is a frequent use case and when done at high throughput, the creation of string copies to capture the dictionary keys would result in excessive time spent in memory allocations.

This PR addresses the issue by using a local allocator in the `byteArrayDictionary` type to amortize the cost of memory allocations for the dictionary keys.


allocator.go
```diff
@@ -0,0 +1,57 @@
...
+	b := a.makeBytes(len(v))
+	copy(b, v)
+	return unsafecast.BytesToString(b)
+}
+
+func (a *allocator) reset() {
+	a.buffer = a.buffer[:0]
+}
+
+// rowAllocator is a memory allocator used to make copy of rows referencing
```
kevinburkesegment, line 41:
```suggestion
// rowAllocator is a memory allocator used to make a copy of rows referencing
```

allocator.go
```diff
@@ -0,0 +1,57 @@
...
+// memory buffers that parquet-go does not have ownership of.
+//
+// This type is used in the implementation of a various readers and writers that
```
kevinburkesegment, line 44:
```suggestion
// This type is used in the implementation of various readers and writers that
```

allocator.go
```diff
@@ -0,0 +1,57 @@
...
+// need to capture rows passed to the ReadRows/WriteRows methods.
+//
+// See: RowBuffer, DedupeRowReader, DedupeRowWriter
+type rowAllocator struct{ allocator }
+
+func (a *rowAllocator) capture(row Row) {
+	for i, v := range row {
+		switch kind := v.Kind(); kind {
+		case ByteArray, FixedLenByteArray:
+			row[i].ptr = unsafecast.AddressOfBytes(a.copyBytes(v.byteArray()))
```
kevinburkesegment, line 54: default case?

allocator.go
```diff
@@ -0,0 +1,57 @@
...
+	copy(b, v)
+	return unsafecast.BytesToString(b)
+}
+
+func (a *allocator) reset() {
+	a.buffer = a.buffer[:0]
+}
+
+// rowAllocator is a memory allocator used to make copy of rows referencing
+// memory buffers that parquet-go does not have ownership of.
```
kevinburkesegment, line 42: I'd appreciate another sentence about why we can't just use the Go allocator for this

allocator.go
achille-roussel: There's nothing to do for the default case 👍

allocator.go
achille-roussel: Let me know how that reads https://github.com/segmentio/parquet-go/pull/445/commits/d2f5a1f9ca5dbfc8a553c8a535a09d00ffc54c87

kevinburkesegment (approved):

PR #450, opened 2022-12-09
Fix readme example

joe-elliott:
The readme example, as written, will call `pages.Close()` before the loop.


achille-roussel (approved):
Thanks 🙇

PR #452, opened 2022-12-09
Transform

achille-roussel:
This PR adds wrappers for `parquet.RowReader` and `parquet.RowWriter` which support applying arbitrary transformations on rows read or written.

PR #453, opened 2022-12-12
Change the memory layout of parquet.Row

achille-roussel:
This PR modifies the in-memory layout of `parquet.Row` to keep values of columns contiguous in memory.

Prior to this change, it was difficult to work with repeated groups materialized in `parquet.Row` because the values of their inner columns would be interleaved. For a parquet schema such as:
```
message {
  repeated group Object {
    string key;
    string value;
  }
}
```
parquet-go would lay out values in `parquet.Row` like this:
```
[key-0, value-0, key-1, value-1, key-2, value-2]
```
With this PR, the layout will be:
```
[key-0, key-1, key-2, value-0, value-1, value-2]
```
All values of each column are now contiguous in the slice, which brings a few improvements:

- the column indexes of values in a `parquet.Row` are now guaranteed to be incrementing, whereas they could have 
- it is now possible to write methods like `parquet.Row.Range` without requiring copying the row to intermediary buffers
- it is possible to deconstruct a `parquet.Row` into a `[][]parquet.Value` and back into a `parquet.Row` without using a schema (and applying a recursive algorithm to walk through the nodes)
- many of the internal algorithms are simplified thanks to this new memory layout (a lot of recursive code has been removed, comparison functions are simpler to write, etc...)
- new conversion rules can be implemented which were difficult to do before

This PR also supersedes and simplifies the transform APIs introduced in https://github.com/segmentio/parquet-go/pull/452

### DIFF
--- a/column_mapping.go
+++ b/column_mapping.go
@@ -52,6 +52,42 @@ func (group columnMappingGroup) lookup(path columnPath) leafColumn {
 	return leafColumn{columnIndex: -1}
 }
 
+func (group columnMappingGroup) lookupClosest(path columnPath) leafColumn {
+	for len(path) > 0 {
+		g, ok := group[path[0]].(columnMappingGroup)
+		if ok {
+			group, path = g, path[1:]
+		} else {
+			firstName := ""
+			firstLeaf := (*columnMappingLeaf)(nil)
+			for name, child := range group {
+				if leaf, ok := child.(*columnMappingLeaf); ok {
+					if firstLeaf == nil || name < firstName {
+						firstName, firstLeaf = name, leaf
+					}
+				}
+			}
+			if firstLeaf != nil {
+				return firstLeaf.column
+			}
+			break
+		}
+	}
+	return leafColumn{columnIndex: -1}
+}
+
+func (group columnMappingGroup) lookupClosestParent(path columnPath) columnPath {
+	for i := 0; i < len(path); i++ {
+		g, ok := group[path[i]].(columnMappingGroup)
+		if ok {
+			group = g
+		} else {
+			return path[:i]
+		}
+	}
+	return nil
+}
+
 type columnMappingLeaf struct {
 	column leafColumn
 }
--- a/column_mapping.go
+++ b/column_mapping.go
@@ -76,18 +76,6 @@ func (group columnMappingGroup) lookupClosest(path columnPath) leafColumn {
 	return leafColumn{columnIndex: -1}
 }
 
-func (group columnMappingGroup) lookupClosestParent(path columnPath) columnPath {
-	for i := 0; i < len(path); i++ {
-		g, ok := group[path[i]].(columnMappingGroup)
-		if ok {
-			group = g
-		} else {
-			return path[:i]
-		}
-	}
-	return nil
-}
-
 type columnMappingLeaf struct {
 	column leafColumn
 }
--- a/convert.go
+++ b/convert.go
@@ -41,7 +41,7 @@ func (e *ConvertError) Error() string {
 type Conversion interface {
 	// Applies the conversion logic on the src row, returning the result
 	// appended to dst.
-	Convert(dst, src Row) (Row, error)
+	Convert(rows []Row) (int, error)
 	// Converts the given column index in the target schema to the original
 	// column index in the source schema of the conversion.
 	Column(int) int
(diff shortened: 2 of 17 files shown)


joe-elliott:
Does this open up the possibility of preallocating slices when constructing the schema object b/c the length would be known beforehand?

column_mapping.go
```diff
@@ -52,6 +52,30 @@ func (group columnMappingGroup) lookup(path columnPath) leafColumn {
 	return leafColumn{columnIndex: -1}
 }
 
+func (group columnMappingGroup) lookupClosest(path columnPath) leafColumn {
+	for len(path) > 0 {
+		g, ok := group[path[0]].(columnMappingGroup)
+		if ok {
+			group, path = g, path[1:]
```
kevinburkesegment, line 59: does this break if you change the function signature to `(group *columnMappingGroup)`?

it's maybe fine it just makes me nervous if someone does that and doesn't notice this change.

convert.go
```diff
@@ -50,171 +50,167 @@ type Conversion interface {
...
-	targetToSourceIndex []int16
-	sourceToTargetIndex []int16
-	convertFunc         convertFunc
-	schema              *Schema
-	buffers             sync.Pool
+	columns []conversionColumn
+	schema  *Schema
+	buffers sync.Pool
+	// This field is used to size the column buffers held in the sync.Pool since
+	// they are intended to the source rows being converted from.
```
kevinburkesegment, line 57: Missing a verb here

page_test.go
```diff
@@ -486,7 +486,7 @@ func TestRepeatedPageTrailingNulls(t *testing.T) {
 	defer reader.Close()
 
 	n, err := reader.ReadRows(rows)
-	if err != io.EOF {
+	if err != nil && err != io.EOF {
```
kevinburkesegment, line 489: I ran into this as well on a different PR and want to make sure this is desired behavior. Should we handle the EOF internally?

transform.go
```diff
@@ -3,100 +3,50 @@ package parquet
...
 //
-// The transformation function writes the transformed src row to dst, returning
-// the number of rows it has written. It is possible for a single row to be
-// transformed to zero or more rows; transforming to zero rows is similar to
-// applying a filter since the row will be skipped. If the dst buffer is not
-// large enough to contain the transformation, the function must return the
-// sentinel error parquet.ErrShortBuffer.
-func TransformRowReader(reader RowReader, transform func(dst []Row, src Row) (int, error)) RowReader {
+// The transformation function appends the transformed src row to dst, returning
+// dst and any error that occured during the transformation. If dst is returned
```
kevinburkesegment, line 7:
```suggestion
// dst and any error that occurred during the transformation. If dst is returned
```

transform.go
```diff
@@ -142,21 +92,17 @@ func (b *transformRowBuffer) len() int {
...
 //
-// The transformation function writes the transformed src row to dst, returning
-// the number of rows it has written. It is possible for a single row to be
-// transformed to zero or more rows; transforming to zero rows is similar to
-// applying a filter since the row will be skipped. If the dst buffer is not
-// large enough to contain the transformation, the function must return the
-// sentinel error parquet.ErrShortBuffer.
-func TransformRowWriter(writer RowWriter, transform func(dst []Row, src Row) (int, error)) RowWriter {
+// The transformation function appends the transformed src row to dst, returning
+// dst and any error that occured during the transformation. If dst is returned
```
kevinburkesegment, line 96:
```suggestion
// dst and any error that occurred during the transformation. If dst is returned
```

value.go
```diff
@@ -611,13 +618,44 @@ func (v Value) Format(w fmt.State, r rune) {
 		case w.Flag('+'):
 			fmt.Fprintf(w, "%+[1]c %+[1]d %+[1]r %+[1]s", v)
 		case w.Flag('#'):
-			fmt.Fprintf(w, "parquet.Value{%+[1]c, %+[1]d, %+[1]r, %+[1]s}", v)
+			v.formatGoString(w)
+			//fmt.Fprintf(w, "parquet.Value{%+[1]c, %+[1]d, %+[1]r, %+[1]s}", v)
```
kevinburkesegment, line 622: Stray?

kevinburkesegment (approved):

column_mapping.go
achille-roussel: It does break, but then unit tests should break since the behavior of the method will change.

page_test.go
achille-roussel: This is the desired behavior, it matches the behavior of `io.Reader`, which may not return `io.EOF` if values were read.

PR #454, opened 2022-12-12
Add File Schema Option

joe-elliott:
In Tempo we have a process that can hold 1000s of Parquet files open in a highly multitenant environment. In this case we can have millions of schema objects on the heap.

This PR adds the ability to pass a schema into the `File` object on `OpenFile`. By passing in a known, shared schema we can drastically reduce the total number of objects on the heap.

image


achille-roussel (approved):
Nice optimization!

PR #456, opened 2022-12-13
Type conversions

achille-roussel:
Based on https://github.com/segmentio/parquet-go/pull/453, this PR adds more conversions between column types, and especially takes advantage of creating logical parquet types to make opinionated decisions about what conversion rules we support (e.g. when the target type is the logical string type, convert the source values to a human-readable representation).

The changes are based on the APIs introduced in #393

In the future we could imagine making these rules configurable, but I think the code here will cover most use cases for now.

I also took the opportunity of this PR to cleanup code in `type.go`.

### DIFF
--- a/convert.go
+++ b/convert.go
@@ -1,10 +1,14 @@
 package parquet
 
 import (
+	"encoding/binary"
 	"fmt"
 	"io"
+	"math"
+	"strconv"
 	"sync"
 
+	"github.com/segmentio/parquet-go/deprecated"
 	"github.com/segmentio/parquet-go/encoding"
 )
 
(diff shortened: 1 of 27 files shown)


convert.go
```diff
@@ -565,3 +573,475 @@ func (c *convertedRows) Schema() *Schema {
...
+	return v.convertToInt64(targetValue), nil
+}
+
+const nanosecondsPerDay = 24 * 60 * 60 * 1e9
+
+func days(t time.Time) int {
+	return int(t.Sub(unixEpoch).Hours()) / 24
+}
+
+func midnight(t time.Time) time.Time {
```
kevinburkesegment, line 1021: Maybe doc that it's midnight _before_ the date and not after?

errors.go
```diff
@@ -54,6 +54,11 @@ var (
 	// ErrTooManyRowGroups is returned when attempting to generate a parquet
 	// file with more than MaxRowGroups row groups.
 	ErrTooManyRowGroups = errors.New("the limit of 32767 row groups has been reached")
+
+	// ErrConversion is used to indicate that a conversion betwen two values
+	// cannot be done because there are no rules to translate between their
+	// physical types.
+	ErrInvalidConversion = errors.New("invalid conversion between parquet values")
```
kevinburkesegment, line 61: If you are hitting this error you would likely want information about what types you are trying to convert between

Ie. I imagine a much more useful error would be 

cannot convert between %q and %q (no rules exist to perform the conversion)

type.go
```diff
@@ -381,12 +401,32 @@ func (t int32Type) AssignValue(dst reflect.Value, src Value) error {
...
 func (t int32Type) ConvertValue(val Value, typ Type) (Value, error) {
-	return val, checkTypeKindEqual(t, typ)
+	switch typ.(type) {
+	case *stringType:
+		return convertStringToInt32(val)
+	}
+	switch typ.Kind() {
+	case Boolean:
+		return convertBooleanToInt32(val)
+	case Int32:
```
kevinburkesegment, line 415: Another option I'm wondering is to throw an error if you try to convert to yourself, to try to prevent people from doing unnecessary conversions (I'm not 100% sure how you are going to use this yet, maybe that's not a feasible option).

type.go
```diff
@@ -1693,83 +1841,59 @@ func (t *timeType) ConvertedType() *deprecated.ConvertedType {
...
+			return convertStringToTimeMicros(val, tz)
+		} else {
+			return convertStringToTimeMillis(val, tz)
+		}
+	case *timestampType:
+		tz := t.tz()
+		if t.Unit.Micros != nil {
+			return convertTimestampToTimeMicros(val, src.Unit, src.tz(), tz)
+		} else {
+			return convertTimestampToTimeMillis(val, src.Unit, src.tz(), tz)
```
kevinburkesegment, line 1893: Maybe just a style thing but I always prefer 

```go
if t.Unit.Micros != nil {
			return convertTimestampToTimeMicros(val, src.Unit, src.tz(), tz)
		}
                 return convertTimestampToTimeMillis(val, src.Unit, src.tz(), tz)
```

which saves you one line and a layer of indent.

kevinburkesegment (approved):

type.go
achille-roussel: I think the alignment of statements help make sense of the code, but that's obviously a very personal opinion.

Let's not aim to follow conventions that a formatting tool isn't going to enforce for us, we'll waste too much time doing a job that a machine would be much better than us at 👍

type.go
achille-roussel: I don't think this is the right layer to enforce it, this API is exported and could be used in contexts where that's perfectly OK to allow conversions between columns of the same type. The code remains correct since the operation is a no-op in this case.

This code is already in use actually since `parquet.Type.ConvertValue` is the underlying mechanism with which `paruqet.Convert` converts columns that change types between two schemas.

errors.go
achille-roussel: This is just the sentinel value to test with `errors.Is`, we always wrap the error where we use it with extra context like what you suggested.

convert.go
achille-roussel: I'll rename the function to be more explicit instead.

PR #457, opened 2022-12-13
add parquet.RowBuilder

achille-roussel:
While it is possible to construct simple `parquet.Row` values _by hand_, in complex cases (e.g. with lots of nesting) managing the get the repetition and definition levels right can be challenging.

With this PR, I'm adding a new type named `parquet.RowBuilder` which helps applications build `parquet.Row` values by having knowledge of the targeted output schema, allowing automatic generation of levels and column indexes, as well as null values where needed.


kevinburkesegment (approved):
This seems like a case where it would be really useful to have an Example or doc.

row_builder.go
```diff
@@ -0,0 +1,185 @@
...
+func (b *RowBuilder) Row() Row {
+	numValues := 0
+	for _, column := range b.columns {
+		numValues += len(column)
+	}
+	return b.AppendRow(make(Row, 0, numValues))
+}
+
+// AppendRow appends the current state of b to row and returns it.
+func (b *RowBuilder) AppendRow(row Row) Row {
```
kevinburkesegment, line 147: FWIW, in the standard library, the Append() functions are static functions that take an existing slice and new value and then return the new slice. Ie here it would be something like

```
func AppendRow(row Row, newContents interface{}(?)) Row

achille-roussel: 🤔 there is [`time.(Time).AppendFormat`](https://pkg.go.dev/time#Time.AppendFormat)

I don't think there's a strong convention being enforced in the stdlib here, but maybe I just don't have enough context on the design discussions that are going on behind the scenes.

PR #462, opened 2022-12-22
Fix SeekToRow on mergedRowGroupRows

asubiotto:
When a Read is performed after SeekToRow on mergedRowGroups, the rowIndex is checked against the seek index and advanced until the rowIndex == seek index. Previously, the rowIndex was not advanced in the normal read path, resulting in mistakenly dropping unread rows when advancing the rowIndex.

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -106,6 +106,12 @@ type mergedRowGroupRows struct {
 	schema    *Schema
 }
 
+func (r *mergedRowGroupRows) readInternal(rows []Row) (int, error) {
+	n, err := r.merge.ReadRows(rows)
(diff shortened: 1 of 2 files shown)


asubiotto:
cc @achille-roussel

asubiotto:
CI should be green now. The issue was returning `n == 0` on `err == io.EOF`.

asubiotto:
Friendly ping @achille-roussel (I apologize if you're still on PTO)

achille-roussel (approved):
Thanks for the reminder! The change looks good 👍

PR #466, opened 2023-01-20
Fix MultiRowGroup pageReadMode

joe-elliott:
Previously we added the ability to configure a "sync" or "async" page read mode:

https://github.com/segmentio/parquet-go/pull/417

This PR corrects an issue where this configuration option is not respected when using MultiRowGroup.


achille-roussel (approved):
Thanks for the fix! 🙌

PR #472, opened 2023-02-06
Enable *int64 to write as optional timestamp, *int32 to write as optional date.

DuanWeiFan:
At the moment, schema does not allow to be set from *int32 to date, or *int64 to timestamp.

We would like to add a statement that will redirect to it's Elem() whenever it encounters a reflect.Ptr.

This change was made in the past, but for some reason it was reverted later on. [here](https://github.com/segmentio/parquet-go/pull/227/files#diff-c3b0d951288715c695f51ce2727c87a056e89ccbb4f3003be71baa8b405383e9)
Adding similar logic back and update schema_test.go


DuanWeiFan:
Hi, can someone help take a look at this PR?
The main purpose is to support writing pointers as Parquet schema.

DuanWeiFan:
Hi @achille-roussel could you help take a look at this pr? 
At the moment parquet SchemaOf does not work for some pointers type. Would like to add this function so that it would work for all different pointers. Thank you.

achille-roussel:
Hello @DuanWeiFan, I left Twilio so I'm currently not part of the maintainer team anymore for parquet-go.

Maybe someone else from the team can pick up the review? @kevinburkesegment @bartleyg

PR #474, opened 2023-02-09
parquet: respect NullsFirst sorting order in Buffer

asubiotto:
This was previously not respected because null ordering was set after the column was created.

cc @achille-roussel


bartleyg (approved):
LGTM! Thanks for your contribution 🥇

achille-roussel (approved):
The assignment of `nullOrdering` was ineffective, maybe we need to bring in a static checker here that would have highlighted the issue.

Thanks for the fix and the test either ways! 🙌

PR #476, opened 2023-02-16
parquet: update ReadRows comment to mention need to copy

### LINKED ISSUE #475: corruption of already read byte values on read of subsequent pages
When reading rows from a file, `ReadRows` only reads up to the smallest number of rows remaining in the current page for each column (https://github.com/segmentio/parquet-go/blob/1d85e81366814efe13170410a86be216a8c9b996/row_group.go#L311-L312). When calling `ReadRows` multiple times (for each page), it seems like the byte values read on previous calls to `ReadRows` are overwritten to point to the values in the new page. I would expect the rows returned from `ReadRows` to be safe for reuse without performing a deep copy.

To reproduce, run the following test:
```
func TestReadRowNoOverwrite(t *testing.T) {
	type model struct {
		Bytes []byte
	}
	var b bytes.Buffer
	w := parquet.NewGenericWriter[model](&b, &parquet.WriterConfig{
		PageBufferSize: 1,
	})

	testBytes := make([][]byte, 2)
	for i := range testBytes {
		testBytes[i] = []byte(fmt.Sprintf("%d", i))
		if _, err := w.Write([]model{{Bytes: testBytes[i]}}); err != nil {
			t.Fatal(err)
		}
	}

	if err := w.Close(); err != nil {
		t.Fatal(err)
	}

	rows := parquet.NewGenericReader[model](bytes.NewReader(b.Bytes()))
	defer rows.Close()
	rowBuf := make([]parquet.Row, rows.NumRows())
	for n := 0; int64(n) < rows.NumRows(); {
		readN, err := rows.ReadRows(rowBuf[n:])
		n += readN
		if err != nil {
			if errors.Is(err, io.EOF) {
				break
			}
			t.Fatal(err)
		}
	}
	for i := range testBytes {
		if !bytes.Equal(testBytes[i], rowBuf[i][0].Bytes()) {
			t.Fatalf(
				"row mismatch at %d, expected %v but found %v",
				i, string(testBytes[i]), string(rowBuf[i][0].Bytes()),
			)
		}
	}
}
```

cc @achille-roussel

asubiotto:
Closes #475

### DIFF
--- a/row.go
+++ b/row.go
@@ -126,8 +126,10 @@ type RowSeeker interface {
 
 // RowReader reads a sequence of parquet rows.
 type RowReader interface {
-	// Read rows from the reader, returning the number of rows read into the
-	// buffer, and any error that occurred.
+	// ReadRows reads rows from the reader, returning the number of rows read
+	// into the buffer, and any error that occurred. Note that the rows read
+	// into the buffer are not safe for reuse after a subsequent call to
+	// ReadRows. Callers that want to reuse rows must copy the rows using Clone.
 	//
 	// When all rows have been read, the reader returns io.EOF to indicate the
 	// end of the sequence. It is valid for the reader to return both a non-zero


bartleyg (approved):
Thanks for the update!

achille-roussel (approved):
Thanks for the update!

PR #477, opened 2023-02-16
parquet: add option to reset SortingWriter with a different sortRowCount

asubiotto:
The motivation for this is to allow callers that know how many rows they want to write to be able to reuse pooled SortingWriters.

This is a proposal for now, I have yet to integrate this into our (Polar Signals') usage, but I foresee us needing something like this. Happy to discuss.

### DIFF
--- a/sorting.go
+++ b/sorting.go
@@ -133,9 +133,14 @@ func (w *SortingWriter[T]) Flush() error {
 }
 
 func (w *SortingWriter[T]) Reset(output io.Writer) {
+	w.ResetWithSortRowCount(output, w.maxRows)
+}
+
+func (w *SortingWriter[T]) ResetWithSortRowCount(output io.Writer, sortRowCount int64) {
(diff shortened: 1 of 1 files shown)


asubiotto:
Another option is to just use `math.MaxInt64` and `Flush` when we're done.

asubiotto:
We're probably not going to use the `SortingWriter` given it's a 30% performance hit vs just writing to a buffer, sorting and copying to a writer so happy to close this PR.

kevinburkesegment:
Maybe we need to document the SortingWriter is supposed to be used for stuff where the working set does not fit in memory? (This is its intended use case).

kevinburkesegment:
Achille also recommends using the RowBuffer (instead of just a buffer) but maybe you are doing that.

asubiotto:
Thanks! Actually we are not using the RowBuffer. Will try it out.

PR #478, opened 2023-02-16
Handle io.EOF errors returned by ReadAt opening files

metalmatze:
Hey, 

we're opening parquet files with an object storage client and not directly via `os.File`. This currently fails. 

The underlying [ReaderAt interface is allowed to return io.EOF](https://pkg.go.dev/io#ReaderAt). Currently, opening a file fails for us because reading the footer the object storage client reads until the end of the file returning io.EOF yet at the same time the data was written to the buffer just fine. 

Now we can fix this in our [FrostDB code](https://github.com/polarsignals/frostdb/pull/351), however, we think that generally speaking FrostDB' object storage client is compliant with the ReaderAt interface. 

What is this project's stance on handling the `io.EOF` here?

### DIFF
--- a/file.go
+++ b/file.go
@@ -3,6 +3,7 @@ package parquet
 import (
 	"bufio"
 	"encoding/binary"
+	"errors"
 	"fmt"
 	"hash/crc32"
 	"io"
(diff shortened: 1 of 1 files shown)


achille-roussel (approved):
Looking good, I agree that we should not push this check to the io.ReaderAt implementation 👍

joe-elliott:
This makes sense to me for reading the footer where you would likely get valid data and an EOF, but do we need this for the column and offset indexes?

achille-roussel:
Maybe we can use a helper function to handle unexpected conditions?
```go
func readAt(r io.ReaderAt, b []byte, off int64) (int, error) {
  n, err := r.ReadAt(b, off)
  if n == len(b) {
    err = nil
  } else {
    switch err {
    case nil:
      err = io.ErrNoProgress
    case io.EOF:
      err = io.ErrUnexpectedEOF
    }
  }
  return n, err
}
```
- we don't make assumptions about the underlying io.ReaderAt implementation
- we silence errors if we read all the bytes we were looking for as they don't impact correctness

kevinburkesegment:

PR #479, opened 2023-02-16
Docs: Correct FileConfig's DefaultReadMode to ReadModeSync

bartleyg:
And add suggested ReadBufferSize for network storage.


achille-roussel (approved):
Thanks for the fixes!

Issue #410, opened 2022-11-17
Question: How can stats get added to row group columns?


rcaught:
I'm generating files with the generic writer, but when I read the meta data for my generated files, every row group column has `ST:[no stats for this column]`. This I believe will lead to very poor predicate searching, etc in other tools.

Its unclear of the steps I would have to take to have these stats produced. Do I need to run everything through a buffered sort? Or is this simply not supported yet?

achille-roussel:
Hello @rcaught 

parquet-go currently does not generate column stats but this could be added to the library.

We always welcome contributions, would you have time to submit a pull request which adds this feature?

PR #485, opened 2023-03-16
Row group statistics

suremarc:
See #410. This PR adds row group statistics, which will enable row-group level predicate pushdown. 

The tests mostly pass with the changes I added, only that the column encodings seem to have a nondeterministic order. I had to build `parquet-tools` from scratch, so I'm not sure if this was just me building the wrong version or something.

### DIFF
--- a/writer_test.go
+++ b/writer_test.go
@@ -101,7 +101,7 @@ var writerTests = []struct {
 		},
 		dump: `row group 0
 --------------------------------------------------------------------------------
-first_name:  BINARY ZSTD DO:4 FPO:55 SZ:90/72/0.80 VC:3 ENC:PLAIN,RLE_DICTIONARY ST:[min: Han, max: Luke, num_nulls not defined]
+first_name:  BINARY ZSTD DO:4 FPO:55 SZ:90/72/0.80 VC:3 ENC:RLE_DICTIONARY,PLAIN ST:[min: Han, max: Luke, num_nulls not defined]
 last_name:   BINARY ZSTD DO:0 FPO:94 SZ:127/121/0.95 VC:3 ENC:DELTA_BYTE_ARRAY ST:[min: Skywalker, max: Solo, num_nulls not defined]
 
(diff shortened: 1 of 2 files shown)


bartleyg (changes requested):
Thanks for the contribution! These tests aren't hooked up in Github Actions b/c no parquet-writer is installed so I had to run them locally. I saw your note about non-deterministic encoding order. Since I can only verify locally for these, if you make the suggested changes I can merge this in!

suremarc:
> Thanks for the contribution! These tests aren't hooked up in Github Actions b/c no parquet-writer is installed so I had to run them locally. I saw your note about non-deterministic encoding order. Since I can only verify locally for these, if you make the suggested changes I can merge this in!

Thanks! I made the requested changes. LMK if anything still looks off.

writer_test.go
```diff
@@ -425,8 +425,8 @@ value 3: R:0 D:0 V:<null>
 		},
 		dump: `row group 0
 --------------------------------------------------------------------------------
-name:   BINARY UNCOMPRESSED DO:4 FPO:49 SZ:73/73/1.00 VC:2 ENC:PLAIN,RLE_DICTIONARY ST:[no stats for this column]
-value:  DOUBLE UNCOMPRESSED DO:0 FPO:77 SZ:39/39/1.00 VC:2 ENC:PLAIN ST:[no stats for this column]
+name:   name:   BINARY UNCOMPRESSED DO:4 FPO:49 SZ:73/73/1.00 VC:2 ENC:RLE_DICTIONARY,PLAIN ST:[min: customer1, max: customer2, num_nulls not defined]
```
bartleyg, line 428: there's an extra name: prepended there but this suggestion should work below:
```suggestion
name:   BINARY UNCOMPRESSED DO:4 FPO:49 SZ:73/73/1.00 VC:2 ENC:RLE_DICTIONARY,PLAIN ST:[min: customer1, max: customer2, num_nulls not defined]
```

bartleyg (changes requested):

bartleyg (approved):

Issue #471, opened 2023-02-01
Missing rows after read columns with multiple pages


alexmatsak:
Hi! 
Seems like there is a bug inside `Reader` or maybe I'm doing something wrong, details:

I'm using this code to read rows from in-mem parquet buffer and it works good until column has only 1 page.
```golang
f, err := parquet.OpenFile(r, size)
if err != nil {
	return nil, err
}

rows := make([]T, f.NumRows())
if _, err := parquet.NewGenericReader[T](f).Read(rows); err != nil {
	return nil, err
}
```

When second page appeared(in my case only for single column, didn't check if multiple), because of this [line](https://github.com/segmentio/parquet-go/blob/main/row_group.go#L349) my slice contains only `1674` elements.
```
$ parquet pages ~/my-awesome-file.parquet

Column: Column1.key_value.value.SomeStruct.SomeField
--------------------------------------------------------------------------------
  page   type  enc  count   avg size   size       rows     nulls   min / max
  0-0    data  _ D  1674    141.38 B   231.128 kB 1664     0
  0-1    data  _ D  23      143.39 B   3.221 kB   23       0

Column: Column2
--------------------------------------------------------------------------------
  page   type  enc  count   avg size   size       rows     nulls   min / max
  0-0    data  _ _  1687    8.00 B     13.180 kB  1687     0
.
.
.
```

Just for testing purpose I tried to switch to this code and on second call it fetched missing 23 rows.
```golang
f, err := parquet.OpenFile(r, size)
if err != nil {
	return nil, err
}

rows := make([]T, f.NumRows())
n, err := reader.Read(rows)
if err != nil {
	return nil, err
}

if n != len(rows) {
	n, err = reader.Read(rows[n:])
	if err != nil {
		return nil, err
	}
}
```
Thanks in advance!

PR #489, opened 2023-03-23
Fix Generic Read methods to read all rows up to length of argument

### LINKED ISSUE #469: [Bug] parquet.readFile only reads 1024 rows
This file (zipped parquet file):
[debug.zip](https://github.com/segmentio/parquet-go/files/10542463/debug.zip) has more than 1024 rows (around 9K).

But if I read it using the following code:

```
type RowType struct {
	Name string `parquet:",optional"`
	Data []byte `parquet:",optional"`
}

rows, err := parquet.ReadFile[RowType](file)
```

and print out the number of rows, I only see that there are 1024 rows.
Tools like the `pandas` library from Python don't have this issue.

bartleyg:
This fixes the generic Read methods where only one page worth of rows was read instead of all rows up to length of the argument slice. This affected at least `parquet.ReadFile`, `parquet.Read`, and `GenericReader.Read`.

You can see a failed test here https://github.com/segmentio/parquet-go/actions/runs/4515595007/jobs/7953065166 without the fix with a new test case using larger sizes here https://github.com/segmentio/parquet-go/pull/489/files#diff-306797de164c962fe5b992da67ab7ff245b8a8e7c76ca21e350ee08c7071d39eR494.

I added a new generic Read path test with multiple RowGroups & multiple Pages to increase the test coverage around this.

I also noted that the TestWriter writerTests have been skipped since using Github Actions because parquet-tools are not installed in Github CI.

Fixes #469, #471, & possibly others.


parquet_go18_test.go
```diff
@@ -326,6 +326,72 @@ func TestIssue377(t *testing.T) {
...
+		}
+		// Flush writes rows as row group. 4 total (20k/5k) in this file.
+		if (i+1)%maxPageBytes == 0 {
+			err = w.Flush()
+			if err != nil {
+				t.Fatal("w.Flush: ", err)
+			}
+		}
+	}
+	err = w.Close()
```
kevinburkesegment, line 368: Can shorten the test and also avoid shadowing this variable by doing e.g.

```go
if err := w.Close(); err != nil {
    t.Fatalf("w.Close: %v", err)

File: reader_go18.go [hunk 2]
```diff
@@ -127,24 +134,40 @@ func (r *GenericReader[T]) Close() error {
 	return r.base.Close()
 }
 
+// readRows reads the next rows from the reader into the given rows slice up to len(rows).
```
kevinburkesegment, line 137: > // readRows reads the next rows from the reader into the given rows slice up to len(rows).

How about 

> // readRows reads the next rows from the reader into r, up to len(rows).

parquet_go18_test.go
bartleyg: It's not shadowed. Doing the suggestion would shadow. I find the `if err := work(); err != nil {` pattern very unreadable just to save a line.

PR #491, opened 2023-03-28
fix for reading less than expected number of rows (#469)

ryandeivert:
resolves: https://github.com/segmentio/parquet-go/issues/469
related to: https://github.com/segmentio/parquet-go/issues/471

## Background

The `Read` convenience method does not do any pagination across row reads, and simply returns the first group to be read in the buffer

## Changes 

Implementing simple "pagination" of rows to properly fill the buffer with expected set of rows

## Testing

Adding unit test and fixture file for testing this with a file of 8000 rows


parquet_go18.go
```diff
@@ -27,12 +27,19 @@ func Read[T any](r io.ReaderAt, size int64, options ...ReaderOption) (rows []T,
...
+	var i int
+	for {
+		c, err := reader.Read(rows[i:])
+		if err == io.EOF {
+			break
+		}
+		if err != nil {
+			return rows, err
+		}
+		i = c
```
grantwwu, line 39: Hi, also been looking for a fix for this. I appreciate that someone else has been working on getting this fixed!

I'm a little concerned about this code - I believe that `reader.Read` returns the number of rows read. Shouldn't we be setting `i += c`?

ryandeivert: hi @grantwwu yes so sorry, good catch. updated just now

grantwwu:
Does the test fail without the fix? In my testing, I've seen `reader.Read` read up to 2^15 rows at once. But that was with rows containing a single int64 (that were probably being stored in a single byte each due to delta encoding).

bartleyg:
@ryandeivert thanks for the contribution! And sorry this was being worked on at a lower level to fix some related issues. Should be fixed by https://github.com/segmentio/parquet-go/pull/489 so pull main and try it out.

ryandeivert:
@bartleyg great thanks!

PR #496, opened 2023-03-31
Increment truncated binary in byteArrayColumnIndexer

ty-sentio-xyz:
Byte array will be increased by 1 after truncated, for both byteArrayColumnIndexer and fixedLenByteArrayColumnIndexer.

This is to make sure truncated values are lexicographically correct, when comparing truncated value with actual max value.


column_index.go
```diff
@@ -676,19 +676,29 @@ func truncateLargeMinByteArrayValue(value []byte, sizeLimit int) []byte {
...
 }
 
-func isMaxByteArrayValue(value []byte) bool {
-	for i := range value {
-		if value[i] != 0xFF {
-			return false
+// incrementByteArray increments the given byte array by 1.
+// Returns nil if the given byte array is all 0xFF.
+// Reference: https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/BinaryTruncator.java#L124
+func IncrementByteArray(value []byte) []byte {
```
bartleyg, line 690: make this private
```suggestion
func incrementByteArray(value []byte) []byte {
```

bartleyg (commented):
Thanks for the report and PR @ty-sentio-xyz! Can you write tests for byteArrayColumnIndexer and fixedLenByteArrayColumnIndexer showing the current incorrect behavior using the public `parquet.` package? Not sure how possible that is but maybe [this](https://github.com/segmentio/parquet-go/blob/39e02a5ed102bcbf0559dc2259f955e2dc088dba/search_test.go#L55-L88) may help?

ty-sentio-xyz:
Thanks for the comments.

I've updated the PR with a new test case that covers the case.

Not sure this is the right way of doing it but I am separating these tests into two files:
- `column_index_test.go`: sits in `parquet_test` package to test public members of `parquet` package.
- `column_index_internal_test.go`: sits in `parquet` package so that private methods can be tested.

The new test case is:
- `column_index_test.go:TestBinaryColumnIndexMinMax`: where we construct a few min-max indexes and searches, to verify the behavior of min-max index for both fixedLenByteArray and byteArray type (and the test would fail if we comment out the part where byte array gets incremented).

column_index.go
```diff
@@ -676,19 +676,29 @@ func truncateLargeMinByteArrayValue(value []byte, sizeLimit int) []byte {
...
+		elem := newValue[i]
+		elem++
+		newValue[i] = elem
+		if elem != 0 { // Did not overflow: 0xFF -> 0x00
+			return newValue
 		}
 	}
-	return true
+	return nil
 }
```
bartleyg, line 702: Looking this over again, we should avoid the unnecessary allocations. And in the rare case that it fully overflowed we just correct it back.:
```
// incrementByteArray increments the given byte array by 1.
// Reference: https://github.com/apache/parquet-mr/blob/master/parquet-column/src/main/java/org/apache/parquet/internal/column/columnindex/BinaryTruncator.java#L124
func incrementByteArray(value []byte) {
	for i := len(value) - 1; i >= 0; i-- {
		value[i]++
		if value[i] != 0 { // Did not overflow: 0xFF -> 0x00
			return
		}
	}
	// Fully overflowed, so restore all to 0xFF
	for i := range value {
		value[i] = 0xFF
	}
}
```

column_index.go
```diff
@@ -676,19 +676,29 @@ func truncateLargeMinByteArrayValue(value []byte, sizeLimit int) []byte {
...
 
 func truncateLargeMaxByteArrayValue(value []byte, sizeLimit int) []byte {
-	if len(value) > sizeLimit && !isMaxByteArrayValue(value) {
-		value = value[:sizeLimit]
+	if len(value) > sizeLimit {
+		if newValue := incrementByteArray(value[:sizeLimit]); newValue != nil {
+			value = newValue
+		}
```
bartleyg, line 682: This changes now that we don't allocate.
```suggestion
		value = value[:sizeLimit]
		incrementByteArray(value)
```

column_index_internal_test.go
```diff
@@ -0,0 +1,24 @@
...
+	"bytes"
+	"testing"
+)
+
+func TestIncrementByteArray(t *testing.T) {
+	testCases := [][]byte{
+		{0x00, 0x01, 0x02, 0x03}, {0x00, 0x01, 0x02, 0x04},
+		{0x00, 0x01, 0x02, 0xFF}, {0x00, 0x01, 0x03, 0x00},
+		{0x00, 0x01, 0xFF, 0xFF}, {0x00, 0x02, 0x00, 0x00},
+		{0xFF, 0xFF, 0xFF, 0xFF}, nil,
```
bartleyg, line 13: this too:
```suggestion
		{0xFF, 0xFF, 0xFF, 0xFF}, {0xFF, 0xFF, 0xFF, 0xFF},
```

column_index_internal_test.go
```diff
@@ -0,0 +1,24 @@
...
+	}
+
+	for i := 0; i < len(testCases); i += 2 {
+		input := testCases[i]
+		expected := testCases[i+1]
+		actual := incrementByteArray(input)
```
bartleyg, line 19: this too:
```suggestion
		actual := copyBytes(input)
		incrementByteArray(actual)
```

bartleyg (changes requested):

column_index.go
ty-sentio-xyz: Hi, thanks for commenting.

I looked into a few places, it seems the call stack would be like this, for ByteArrayColumn:
- [byteArrayColumnIndexer.IndexPage](https://github.com/segmentio/parquet-go/blob/main/column_index.go#L508): did not copy min, `min.byteArray()` is a cast
- [writerColumn.recordPageStats](https://github.com/segmentio/parquet-go/blob/main/writer.go#L1313): did not copy returned values from `page.Bounds()`
- [byteArrayPage.Bounds](https://github.com/segmentio/parquet-go/blob/main/page.go#L1037): using returned values from `page.bounds()` without copying, where returned values of `page.bounds()` are taken from `page.values` by reslicing, so no copy as well.

If I am not mistaken, for byteArrayColumnIndex, modifying the input slice in place would change the underlying buffer of `page.values` as well.  

If we want to eliminate memory alloc, I'd propose we do it like the way we deal with FixedLenByteArrayColumn, that is appending to `byteArrayColumnIndexer.minValues` and `byteArrayColumnIndexer.maxValues` first (which is effectively a copy), then modify that buffer in place.

PR #503, opened 2023-04-27
Fix #501

### LINKED ISSUE #501: Library sometimes flips boolean values
Somewhat regularly parquet-go will flip boolean values. The test at the bottom reproduces this issue.

Some things I've noticed:
- the issue occurs even if i test with the `purego` tag
- if i don't call `w.Flush()` in the write loop and allow `w.Close()` to flush the issue does not occur
- if i don't randomize the bool set to true the issue does not occur

```
type Bools struct {
	A bool
	B bool
}

func TestBools(t *testing.T) {
	f, err := os.CreateTemp(t.TempDir(), "data.parquet")
	if err != nil {
		t.Fatal(err)
	}
	defer f.Close()
	w := parquet.NewGenericWriter[*Bools](f)

	// write some data
	for i := 0; i < 100; i++ {
		// set one random bool true
		st := &Bools{}
		n := rand.Intn(2)
		switch n {
		case 0:
			st.A = true
		case 1:
			st.B = true
		}
		w.Write([]*Bools{st})
		if i%10 == 0 {
			w.Flush()
		}
	}
	w.Close()

	stat, err := f.Stat()
	if err != nil {
		t.Fatal(err)
	}

	pf, err := parquet.OpenFile(f, stat.Size())
	if err != nil {
		t.Fatal(err)
	}

	r := parquet.NewGenericReader[*Bools](pf)
	for {
		sts := make([]*Bools, 1)
		_, err := r.Read(sts)
		st := sts[0]
		if st != nil {
			if st.A && st.B {
				t.Fatal("both true")
			}
		}
		if err == io.EOF {
			break
		}
		if err != nil {
			t.Fatal(err)
		}
	}
}
```

joe-elliott:
This PR aims to fix #501. I'm not 100% confident in this fix, but based on my understanding of the code in question this is valid.

The issue appears to be this line here:

https://github.com/segmentio/parquet-go/blob/main/column_buffer.go#L851

The code builds a bitmask and assigns the value to `b` to set relevant booleans in the current bit. Then it `|=`s the bitmask against the current bit. Interestingly using `|=` guarantees that this path can never drop an existing bit flag to 0 and whatever happened to be leftover in the buffer persists.

Currently running the Tempo test suite against this.

### DIFF
--- a/column_buffer.go
+++ b/column_buffer.go
@@ -848,7 +848,7 @@ func (col *booleanColumnBuffer) writeValues(rows sparse.Array, _ columnLevels) {
 			}
 			x := uint(col.numValues) / 8
 			y := uint(col.numValues) % 8
-			col.bits[x] |= (b << y) | (col.bits[x] & ^(0xFF << y))
+			col.bits[x] = (b << y) | (col.bits[x] & ^(0xFF << y))
 			col.numValues += int32(i)
 		}
 
--- a/column_buffer_test.go
+++ b/column_buffer_test.go
@@ -1,6 +1,8 @@
 package parquet
 
-import "testing"
+import (
+	"testing"
+)
 
 func TestBroadcastValueInt32(t *testing.T) {
 	buf := make([]int32, 123)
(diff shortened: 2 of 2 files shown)


bartleyg (approved):
LGTM thanks @joe-elliott !

PR #506, opened 2023-05-11
Support reading int96 as timestamp

chelseajonesr:
By default Spark writes parquet files with int96 timestamps, even though the type has been deprecated for a while.  There are details of the format and some discussion here:
https://github.com/apache/parquet-format/pull/49

This PR adds support for reading int96 values into timestamps.  (Just reading, not writing.)

During implementation I found that the `convertToType()` function in `convert.go` was converting in the wrong direction.  I've added a test scenario "string to int" in `convert_test.go` which fails with the existing code and succeeds with the updated version.
Also, converting types with null values failed in some situations, so I added null checks to the `ConvertValue` functions in `type.go` and added the "nils" scenario in `convert_test.go`.

### DIFF
--- a/convert.go
+++ b/convert.go
@@ -83,7 +83,7 @@ func convertToSelf(column []Value) error { return nil }
 func convertToType(targetType, sourceType Type) conversionFunc {
 	return func(column []Value) error {
 		for i, v := range column {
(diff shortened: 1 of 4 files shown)


kevinburkesegment:

PR #509, opened 2023-06-23
Reuse reconstruction columns slice to reduce allocations

zolstein:
Create an internal method on Schema to reconstruct a value, passing in a [][]Value to use as columns. Use this internal method, rather than Reconstruct, when reading rows in GenericReader.

Calling Reconstruct on every row being read, and constructing a new [][]Value, in aggregate, accounts for the majority of allocations while reading parquet files and induces an unnecessarily large GC overhead.

-----

<details>

<summary>Test code used to identify issues</summary>

test.parquet is a 768MB parquet file with 32M records.

```go
type TestStruct struct {
	Field1 int64
	Field2 int64
	Field3 int64
}

func main() {
	entries := make([]TestStruct, 1024)
	inFile, err := os.Open("test.parquet")
	if err != nil {
		log.Fatalf("failed to open parquet file: %v", err)
	}
	pr := parquet.NewGenericReader[TestStruct](inFile)
	for {
		_, err := pr.Read(entries)
		if err == io.EOF {
			break
		} else if err != nil {
			log.Fatalf("failed to read parquet entries: %v\n", err)
		}
	}

	f, err := os.Create("mem.pprof")
	if err != nil {
		log.Fatalf("failed to open file: %v", err)
	}
	defer f.Close()
	if err := pprof.Lookup("allocs").WriteTo(f, 0); err != nil {
		log.Fatalf("failed to write heap profile: %v", err)
	}
}
```

</details>

<details>

<summary>Profile output (before change)</summary>

profile016

</details>

<details>

<summary>Profile output (after change)</summary>

profile014

</details>

### DIFF
--- a/reader_go18.go
+++ b/reader_go18.go
@@ -149,6 +149,9 @@ func (r *GenericReader[T]) readRows(rows []T) (int, error) {
 		r.base.rowbuf = r.base.rowbuf[:nRequest]
 	}
 
+	schema := r.base.Schema()
+	columns := make([][]Value, len(schema.columns))
+
 	var n, nTotal int
 	var err error
 	for {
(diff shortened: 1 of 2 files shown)


reader_go18.go
```diff
@@ -149,6 +149,9 @@ func (r *GenericReader[T]) readRows(rows []T) (int, error) {
 		r.base.rowbuf = r.base.rowbuf[:nRequest]
 	}
 
+	schema := r.base.Schema()
+	columns := make([][]Value, len(schema.columns))
```
zolstein-clumio, line 153: Arguably these columns should be stored in the GenericReader object and reused across Read calls. That would help if Read is called many times with a relatively small rows slice. However, it's harder to know for sure that won't cause problems.

kevinburkesegment:

